PR #10, opened 2022-04-20
[liquid] change claim tx size

sputn1ck:
The new tx size is the correct __virtual__ byte size of the claim transaction


wtogami:
tested ACK

PR #11, opened 2022-04-21
[CLN] Log panics to file (temporary fix)

sputn1ck:
Go panics are written to Stderr, which CLN does not catch.

This commit creates a file in `.lightning/<network>/peerswap` to which all panics are written to.

### DIFF
--- a/clightning/clightning_dev.go
+++ b/clightning/clightning_dev.go
@@ -20,7 +20,40 @@ var (
 )
 
 func init() {
-	devmethods = append(devmethods, &FaucetMethod{}, &GenerateMethod{}, &BigInvoice{}, &BigPay{})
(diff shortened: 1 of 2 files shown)


wtogami:
Please note somewhere that this is a temporary solution until we figure out how to dump backtraces to CLN's log.

Where does LND CLN log to btw?

sputn1ck:
> Please note somewhere that this is a temporary solution until we figure out how to dump backtraces to CLN's log.

This will prob. be a permanent solution as CLN will not capture stderr, but pass it on to the process starting the cln daemon

> Where does LND PeerSwap log to btw?

In peerswapd panics are shown in journal and in the file

wtogami:
> This will prob. be a permanent solution as CLN will not capture stderr, but pass it on to the process starting the cln daemon

There must be a solution possible where either stderr is logged by CLN, or we are able to tell golang to dump the backtrace to CLN's logger function. Let's consult with the CLN team.

wtogami:
Add `peerswap-dev-debugcrash`?

PR #34, opened 2022-05-03
check onchain balance on swap-out request

sputn1ck:
adds a balance check on an incoming swap-out request

### DIFF
--- a/swap/actions.go
+++ b/swap/actions.go
@@ -368,6 +368,19 @@ func (c *CreateSwapOutFromRequestAction) Execute(services *SwapServices, swap *S
 		return swap.HandleError(err)
 	}
 
+	// Check if onchain balance is sufficient for swap + fees + some safety net
+	walletBalance, err := wallet.GetOnchainBalance()
+	if err != nil {
+		return swap.HandleError(err)
+	}
+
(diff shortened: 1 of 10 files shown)


swap/actions.go
```diff
@@ -368,6 +368,16 @@ func (c *CreateSwapOutFromRequestAction) Execute(services *SwapServices, swap *S
 		return swap.HandleError(err)
 	}
 
+	// Check if onchain balance is sufficient for swap + fees + some safety net
+	walletBalance, err := wallet.GetOnchainBalance()
+	if err != nil {
+		return swap.HandleError(err)
+	}
+
+	if walletBalance < swap.GetAmount()+openingFee+20000 {
```
wtogami, line 377: Fine for now but maybe add TODO to comment that this safety buffer be calculated to be more precise. Otherwise sometimes people will wonder why a swap is failing when they think they have enough.

wtogami:
To properly test this, is it required to patch the swap provider side alone, or also the requester?

sputn1ck:
> To properly test this, is it required to patch the swap provider side alone, or also the requester?

Only the provider side needs to be updated

PR #39, opened 2022-05-03
comment out potential source of error

niftynei:
Updated to latest master, failed to start with this error being written
to the crash logs.

Commenting out this block "fixed" the problem.

```
2022/05/03 09:21:29 [INFO] plugin quitting, error: Method already registered
```


sputn1ck:
Thanks for noticing. The crash only happend on build with 'dev' tag.

Should be fixed in #40

PR #42, opened 2022-05-04
revise README.md, upgrade.md, usage.md

wtogami:
Please review.

Regarding docs/usage.md, it needs a lot more work but this is good enough for release.

For now I removed mention of "In order to swap-out you need a minimum balance of liquid bitcoin in order to pay for transaction fees." because it is confusing and requires a larger rewrite to explain in detail.


nepet (approved):
SGTM

PR #43, opened 2022-05-04
improve swap printing

sputn1ck:
- adds more control over when to print swap information
- allows specific printing of cases such as "canceling after prepayment"

### DIFF
--- a/swap/swap.go
+++ b/swap/swap.go
@@ -382,7 +382,7 @@ func (s *SwapData) ToPrettyPrint() *PrettyPrintSwapData {
 		PeerNodeId:      s.PeerNodeId,
 		Amount:          amount,
 		ShortChannelId:  scid,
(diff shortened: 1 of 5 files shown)


swap/swap.go
```diff
@@ -382,7 +382,7 @@ func (s *SwapData) ToPrettyPrint() *PrettyPrintSwapData {
 		PeerNodeId:      s.PeerNodeId,
 		Amount:          amount,
 		ShortChannelId:  scid,
-		OpeningTxId:     s.OpeningTxBroadcasted.TxId,
+		OpeningTxId:     s.GetOpeningTxId(),
```
nepet, line 385: Is this something we missed earlier?

sputn1ck: yes, actually wanted to do that in #44  but appararently I put that here

nepet (approved):
lgtm

PR #44, opened 2022-05-04
cln swap-in fix help message on invalid input

sputn1ck:
This PR changes the order of error message in order to correcly show the missing parameters


nepet (approved):
lgtm

PR #52, opened 2022-05-15
Fix high cpu load

nepet:
Seems the tx watcher was the problem here. We need to wait longer between the calls to the bitcoin backend in
order to avoid high cpu loads.

### DIFF
--- a/txwatcher/rpctxwatcher.go
+++ b/txwatcher/rpctxwatcher.go
@@ -104,7 +104,7 @@ func (s *BlockchainRpcTxWatcher) StartWatchingTxs() error {
 					return err
 				}
 			default:
-				time.Sleep(time.Millisecond)
+				time.Sleep(100 * time.Millisecond)
 			}
 		}
 	}()


sputn1ck:
Is the unusual cpu usage gone completely? The cpu usage on my peerswap lnd node is also unusually high and it is not using the 'rpctxwatcher' services

wtogami:
I didn't test anything else yet but this reduced my `peerswap-plugin` CPU load from ~12% to almost nothing.

nepet:
> Is the unusual cpu usage gone completely? The cpu usage on my peerswap lnd node is also unusually high and it is not using the 'rpctxwatcher' services

Peerswapd has an idle cpu usage of near 0% on my machine.

sputn1ck:
> > Is the unusual cpu usage gone completely? The cpu usage on my peerswap lnd node is also unusually high and it is not using the 'rpctxwatcher' services
> 
> 
> 
> Peerswapd has an idle cpu usage of near 0% on my machine.

Cool, then we only need to find the peerswapd problem

PR #55, opened 2022-05-19
Fix broken MPPs for large cln payments

nepet:
This PR fixes an issue with large swaps where the cln node is the payer. If a payment exceeds the size of 2^32 msat we need to split it up according to cln MPP rules. One of these rules is, that every payment part has to have the exact same size in msat. This PR fixes the peerswap MPP splitter that used to divide in fixed parts with a rest.

Also adds some tests for MPPs.

Please do not merge directly after review, I will squash the fixups first!

### DIFF
--- a/clightning/clightning.go
+++ b/clightning/clightning.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	log2 "log"
+	"math"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


sputn1ck (approved):
lgtm

Nice idea with always splitting the parts in 10. A `LndCln` MPP test could also be great, just to be sure that the current implementation works

laanwj:
Tested ACK

nepet:
I will merge this now as it is fixing for at least normal channel sizes. I Will provide an other PR for large channels.

PR #62, opened 2022-05-31
Fix clone command in CLN and LND setup guides.

nlflint:
Summary of changes:
1. Changes CLN setup guide to clone from `ElementsProject` repo instead of `sputn1k`.
1. Changes both guides to clone using HTTPS instead of SSH.

Using HTTPS works in more situations since a user doesn't need to setup an Github user and SSH key prior to cloning.


wtogami:
ACK

Note if you're updating only docs add `[skip ci]` to at the end of your commit message so the tests don't need to run.

PR #69, opened 2022-06-01
CLN: INFO print version during startup instead of DEBUG

wtogami:
This should be in everyone's log so it's easier to know what version they are running for bug reports.


nepet (approved):
ACK

Issue #76, opened 2022-06-06
make lnd-install fails with "no rule to make target"


dannydeezy:
In the LND installation instructions, `make lnd-install` fails with:
```
make: *** No rule to make target 'lnd-install'.  Stop.
```

Should the command be `make lnd-release` instead? (looking at Makefile)

nepet:
Fixed #127

PR #77, opened 2022-06-06
Fix make command for lnd setup doc

dannydeezy:
See issue: https://github.com/ElementsProject/peerswap/issues/76

### DIFF
--- a/docs/setup_lnd.md
+++ b/docs/setup_lnd.md
@@ -17,7 +17,7 @@ Clone the peerswap repository and build the peerswap plugin
 ```bash
 git clone https://github.com/ElementsProject/peerswap.git && \
 cd peerswap && \
-make lnd-install
+make lnd-release
 ```
 
 This will install `peerswapd` and `pscli` to your go path


docs/setup_lnd.md
```diff
@@ -17,7 +17,7 @@ Clone the peerswap repository and build the peerswap plugin
 ```bash
 git clone https://github.com/ElementsProject/peerswap.git && \
 cd peerswap && \
-make lnd-install
+make lnd-release
 ```
 
 This will install `peerswapd` and `pscli` to your go path
```
nepet, line 23: The docs state that the files will be installed into the users go path, but looking at the makefile this does not seem to be true. The files are just built in the peerswap root directory. So maybe also change this line to match what this command really does, and/or add a small info on how to move the files to the GOPATH?

nepet (changes requested):

docs/setup_lnd.md
openoms: Can confirm that `peerswapd` and `pscli` is installed in the current `peerswap` directory.
Instead of the GOPATH I move them to a directory in the PATH like `/usr/local/bin/`

nepet:
Has been replaced by #119

Issue #60, opened 2022-05-25
Add test for large payments from lnd


nepet:
Currently we only test MPPs for cln. We also need to test large payments and MPPs that are made by lnd.

PR #78, opened 2022-06-22
Add tests for large swaps

nepet:
This PR changes all integration tests to test for swaps with size 10btc. Swaps of this size need large (wumbo) channels.


nepet:
resolves #60

wtogami:
ACK looks good although I didn't try it locally.

(This reminds me ... we should require wumbo to be enabled in the accompanying CLN or LND daemon? We want fewer variations of configurations to support. CLN `listconfigs` can show if wumbo is enabled. Does LND have a way to query if wumbo is enabled during runtime? This belongs in another ticket if you agree.)

wtogami (approved):

PR #79, opened 2022-06-24
Rename GetSatPerKW function. Fix LND fee unit conversion

huilder:
- Renamed the GetFeePerKw function to GetSatsPerVByte so that the function name corresponds to the units returned.

- Fixed the conversion from sats/kw to sats/vByte for LND's fee estimator. The scaling factor of 4 was applied inverted (1/4) 

The tests were run successfully, however, there does not appear to be test coverage of this area.

### DIFF
--- a/lnd/lnd_wallet.go
+++ b/lnd/lnd_wallet.go
@@ -256,11 +256,11 @@ func NewLndFeeEstimator(ctx context.Context, walletkit walletrpc.WalletKitClient
 	return &LndFeeEstimator{ctx: ctx, walletkit: walletkit}
 }
 
-func (l *LndFeeEstimator) GetFeePerKw(targetBlocks uint32) (float64, error) {
+func (l *LndFeeEstimator) GetSatsPerVByte(targetBlocks uint32) (float64, error) {
 	res, err := l.walletkit.EstimateFee(l.ctx, &walletrpc.EstimateFeeRequest{ConfTarget: int32(targetBlocks)})
 	if err != nil {
 		return 0, err
 	}
 
-	return float64(res.SatPerKw / 4000), nil
+	return float64(res.SatPerKw / 250), nil
 }
--- a/clightning/clightning_wallet.go
+++ b/clightning/clightning_wallet.go
@@ -220,7 +220,7 @@ func (cl *ClightningClient) GetFlatSwapOutFee() (uint64, error) {
 	return cl.bitcoinChain.GetFee(218)
 }
 
-func (cl *ClightningClient) GetFeePerKw(targetblocks uint32) (float64, error) {
+func (cl *ClightningClient) GetSatsPerVByte(targetblocks uint32) (float64, error) {
 	if cl.bitcoinNetwork == &chaincfg.RegressionNetParams {
 		return 1, nil
 	}
--- a/onchain/bitcoin.go
+++ b/onchain/bitcoin.go
@@ -25,7 +25,7 @@ type BitcoinOnChain struct {
 }
 
 type FeeEstimator interface {
-	GetFeePerKw(targetBlocks uint32) (float64, error)
+	GetSatsPerVByte(targetBlocks uint32) (float64, error)
 }
 
 func NewBitcoinOnChain(estimator FeeEstimator, chain *chaincfg.Params) *BitcoinOnChain {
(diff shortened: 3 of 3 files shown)


nepet (changes requested):
Good catch on the wrong unit conversion.
Will approve when the typo character (last char) in the commit message is removed.

huilder:
Aha there are some tests I hadn't seen, I'll try to run them locally and find out what's going wrong

huilder:
I've removed the bad character from the commit msg. Those tests pass locally so maybe the CI failure was a timing issue?

nepet:
Seems the CI failed with the same error again I will trigger it once more and investigate if it fails again.

wtogami:
I re-ran that test and it failed in the same way.

nepet:
I will try it out locally this weekend

wtogami:
Hi @huilder could you please join us in PeerSwap Discord? We are discussing this and a possibly related LND fee estimator bug where sometimes the invoice sent from LND PeerSwap has a blank amount.

https://discord.gg/dvyMPMpm

lnd/lnd_wallet.go
```diff
@@ -256,11 +256,11 @@ func NewLndFeeEstimator(ctx context.Context, walletkit walletrpc.WalletKitClient
...
 
-func (l *LndFeeEstimator) GetFeePerKw(targetBlocks uint32) (float64, error) {
+func (l *LndFeeEstimator) GetSatsPerVByte(targetBlocks uint32) (float64, error) {
 	res, err := l.walletkit.EstimateFee(l.ctx, &walletrpc.EstimateFeeRequest{ConfTarget: int32(targetBlocks)})
 	if err != nil {
 		return 0, err
 	}
 
-	return float64(res.SatPerKw / 4000), nil
+	return float64(res.SatPerKw / 250), nil
```
kornpow, line 265: I think this should be 253 and not 250
https://github.com/lightninglabs/pool/blob/master/rpcserver.go#L684

kornpow: Or maybe it should stay 250, so its always slightly higher than 1 sat/vbyte

huilder: This is just the inverse of the scale factor so `1000 / 4` https://github.com/lightningnetwork/lnd/blob/262591c3331c2170697fe89b8c49ab857db0adf7/lnwallet/chainfee/rates.go#L32

I would use `min(1.1, satsPerVByte)` afterwards to set the floor slightly higher if that is desired.

Also it might be better to use weight units throughout instead of sats/vbyte anyway - it seems to me that kwu is the canonical measure and the vbyte measure is an alternative that serves mostly to keep the bitcoincore fee rpc calls consistent

wtogami:
Discussion in Discord indicates this PR improved the situation but we're still sometimes seeing fee rates below 1.0 sats/vbyte suggesting the calculation method is not entirely fixed.

wtogami:
> We now just have to agree if we have to divide by 250 or 253. It would be good if we have a fast decision on this.

This shouldn't be a matter of decision. Can a little test program synthetically verify the possible fee estimates you can receive from LND and the corresponding sats/perKB fee rate? 

image

@kornpow how did get this repl?

After we confirm 250 or 253 I want to floor the sats/vByte to 1.1 for both CLN and LND.

grubles:
Successfully tested a wumbo-sized swap (2 signet BTC) between a CLN v0.11.2 release node and LND 0.15.0-beta.rc6 node using `res.SatPerKw / 250`.

kornpow:
For sure, the minimum that LNDs EstimateFee gives is 253.
https://github.com/lightningnetwork/lnd/blob/master/lnwallet/chainfee/rates.go#L11-L19
I still dont know whether you would like 250 or 253 in peerswaps code. Either way having a minimum set to 1.1 should be the better solution anyways than agonizing about this constant.

Another useful file:
https://github.com/lightningnetwork/lnd/blob/master/lnwallet/chainfee/estimator.go#L97

I added to this PR and put in a 1.1 sat/vbyte minimum.
https://github.com/ElementsProject/peerswap/pull/100

lnd/lnd_wallet.go
kornpow: Sorry I just responded to this now. I went looking for this file in LND just a bit ago. 250 sounds right to me.

kornpow:
We've gone back and forth, could it make sense to keep it in weight units?
https://mempool.space/tx/2e8065e095a7291aef31a77f8ac727dfdcbdf9187c136f5ac68c49cbe62eea3b
image

nepet:
This PR and its commit have been absorbed into #102.
Feel free to continue the conversation there ^^ :)

Issue #73, opened 2022-06-03
CLN: peerswap-plugin crash after peer failed to pay swap invoice


wtogami:
CLN v0.10.2 PeerSwap 2833bdb8687f5c5b4337449bfb6546f028348ba7
LND v0.14.1-beta PeerSwap 1f0e91fe95b5e8124d035628cb91623fea462b50

LND requested lbtc swap-out from CLN. Swap payment failed to find a route and we know it was the LND node's fault. How?

https://github.com/accumulator/charge-lnd
* They were running charge-lnd which had dropped max_htlc to 1 as part of an automated forwarding discouragement strategy for an imbalanced channel.
* They had ~200k sats on their side so they thought they would do a 100k sats `swap-out` test. They forgot about charge-lnd.
* Since the maximum htlc size is set to 1 sat the fee prepayment invoice was paid via an unrelated route. See: #61 where I argued fee prepayments should be explicitly over the same channel partly as a sanity check.
* The swap payment fails to find a route which is expected in this weird scenario.
* The unexpected part is it causes CLN peerswap-plugin to crash.

```
2022-06-03T08:40:42.487Z DEBUG   plugin-peerswap-plugin: [FSM] event:id: 6a0fbd775103b48e08ac7bf3e873637cc539442ee1a64d05710fcc62e4a37560, Event_OnSwapOutRequestReceived on
2022-06-03T08:40:42.749Z DEBUG   plugin-peerswap-plugin: [FSM] event:id: 6a0fbd775103b48e08ac7bf3e873637cc539442ee1a64d05710fcc62e4a37560, Event_ActionSucceeded on State_SwapOutReceiver_CreateSwap
2022-06-03T08:40:42.749Z INFO    plugin-peerswap-plugin: [Swap:6a0fbd775103b48e08ac7bf3e873637cc539442ee1a64d05710fcc62e4a37560] swap out request received: peer: REDACTEDLNDPUBKEY chanId: REDACTEDHEIGHT:REDACTEDINDEX:1 initiator: REDACTEDLNDPUBKEY amount 100000
2022-06-03T08:40:42.752Z DEBUG   plugin-peerswap-plugin: [FSM] event:id: 6a0fbd775103b48e08ac7bf3e873637cc539442ee1a64d05710fcc62e4a37560, Event_ActionSucceeded on State_SwapOutReceiver_SendFeeInvoice
[...]

PR #83, opened 2022-06-30
Use fix payment route for fee invoice

### LINKED ISSUE #61: Pay fee prepayment via explicit route
```
INFO    plugin-pay: cmd 14102 partid 0: Payment fee constraint 675msat is below exemption threshold, allowing a maximum fee of 5000msat
```

Why?
* This CLN `pay` log print is irrelevant because there is no fee for paying your peer.
* It is beneficial for the pre reimbursement to explicitly happen via the same channel because it doubles as a sanity check that the channel works properly before a possibly costly onchain payment + recovery cost.

nepet:
This pr adds an explicit route (the direct channel to the peer that is swapped on) to the fee payment. Resolves #61

### DIFF
--- a/clightning/clightning.go
+++ b/clightning/clightning.go
@@ -268,6 +268,28 @@ func (cl *ClightningClient) PayInvoice(payreq string) (preimage string, err erro
 	return res.PaymentPreimage, nil
 }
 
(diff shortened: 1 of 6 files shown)


wtogami:
* Note #73 will likely not be easily reproducible after this is applied.
* Please check the test failure.

PR #90, opened 2022-07-06
Update signetguide_lnd.md

grubles:
Add note about including the proper tags if building LND from source.


wtogami:
Fine for now but in the future if your commits are doc only you can add `[skip ci]` as the tests don't need to run if the code had not changed.

ACK

PR #91, opened 2022-07-07
Update cln and lnd versions

nepet:
- cln v0.11.1
- lnd v0.15.0-beta

These are the latest versions available as nix packages.


wtogami:
utACK

PR #97, opened 2022-07-11
Test for basic recovery

nepet:
This pull request addresses some issues where the recovery of a swap (in csv condition) failed with a panic. It adds the following:
 - A basic test for recovery
 - Wait for the node to be in sync with the blockchain before attempting to recover
 - Add some helper functions to the test framework
 - Fixes an issue with the transaction watcher having to wait an extra confirmation


nepet:
To be clear, This pr does not fix the panic we could not reproduce but it encountered some problems that also came up when discussing #73.

cmd/peerswap-plugin/main.go
```diff
@@ -107,6 +108,22 @@ func run() error {
...
+	log.Infof("Waiting for cln to be synced")
+	err = utils.WaitFor(func() bool {
+		info, err := lightningPlugin.GetLightningRpc().GetInfo()
+		if err != nil {
+			log.Infof("rpc.GetInfo() %v", err)
+			return false
+		}
+
+		return info.IsBitcoindSync() && info.IsLightningdSync()
+	}, 10*time.Second, 10*time.Minute)
```
wtogami, line 122:
```
2022-07-13T16:54:45.513Z INFO    plugin-peerswap-plugin: PeerSwap Initialized, running PeerSwap commit 95eac6a5281fb3cfbae549222ebd1aa29485f0b4
2022-07-13T16:54:45.555Z INFO    plugin-peerswap-plugin: Waiting for cln to be synced
2022-07-13T17:04:48.800Z INFO    plugin-peerswap-plugin: Killing plugin: exited during normal operation
```

We found 10 minutes can be too short. Perhaps it should wait forever? It is CLN's problem to let us know when it thinks it is ready.

wtogami (changes requested):

wtogami:
> Wait for the node to be in sync with the blockchain before attempting to recover

This improved things a bit but it but we need more startup checks. PeerSwap gives up too quickly sending things into cancel when it should have seen it is temporary.

For now this is good enough for merge. Want to squash fixups first?

wtogami (approved):

nepet:
Yes, let me squash that fixups!

nepet:
Squashed the fixups and rebased this ugly pr onto master.

wtogami:
ACK 688ebca and tested for several days by a few people 
confirmed 6135d8d is equivalent

PR #104, opened 2022-07-13
Updated default regtest port to the correct port 18443.

cmdruid:
Default regtest port is incorrect. Should be 18443.


wtogami:
@grubles please check if he's updating the correct port?

wtogami:
OK so close not a bug?

grubles:
No, @cmdruid is correct. The regtest port needs to be changed to 18443. Currently, it's 18332, which is what testnet uses.

wtogami:
oops. OK. He needs to squash his commits to a single commit and put the main description in the first line of the commit.

wtogami (changes requested):
Needs squash.
Please put "Updated default regtest port to the correct port 18443." on the first line of the commit message.

PR #106, opened 2022-07-14
update usage doc

grubles:
added LND-specific swap commands and labeled the CLN ones


nepet (approved):
Ack fcd859cd21657c9a0de0c30c22e5d5f6fa47ff48

PR #111, opened 2022-07-25
Add grpc retry to requests and streams

nepet:
This PR adds the go-grpc-middleware `retry`. We need to be able to retry a call or to resubscribe to a stream in the case that the lnd node is unreachable for some time.

This alone will not be sufficient to fix all connection problems with lnd, therefore the watchers (peers, chain, invoices) need a complete rework.
I will provide another PR that takes care of the watcher reworks to keep this PR single tasked and reviewable. That PR will be based on the head of this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,13 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        test-vector: [bitcoin-cln, bitcoin-lnd, liquid-cln, liquid-lnd]
+        test-vector: [
+          bitcoin-cln,
+          bitcoin-lnd,
+          liquid-cln,
+          liquid-lnd,
+          misc-integration
+          ]
     steps:
       - uses: actions/checkout@v2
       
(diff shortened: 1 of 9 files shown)


wtogami:
Testers: Please test like normal. Need to know this at least does not cause regressions.

kornpow:
I did some testing of this PR:
I was able to swap out successfully running LND 0.15.0
I was able to swap in using LBTC, and
I did another swap out to BTC, and while the onchain transaction was waiting to confirm, I turned off my LND node for 5 minutes or so. The logs for the retry/reconnect is really spammy though, it might be better to retry every second or two, instead of many times per second.

wtogami:
We had planned on addressing the separate problem of retries being too fast in another PR. Maybe for now we should add a one second minimum delay to make it less noisy?

kornpow:
Making the logs less noisy might be a good idea, but either in this PR or another is fine.

To follow up on the second swapout above where I restarted LND midway. I noticed the onchain BTC transaction had 8 confirmations, so I restarted `peerswapd` and it successfully finished the swap.

So while it did reconnect to LND, it still is missing some of the subscriptions reconnecting... Which I think is the expected result of this PR. No regressions, but it didn't fix the issue.

wtogami:
gRPC reconnection handling, especially LND restart isn't complete yet. The testing here is only "is this worse than previously". We'll ask for comprehensive reconnection testing when that part is ready.

wtogami:
ACK d76c393259f80d527beb5816afc938eba8ac3241

I'm relying mainly on testing reports from our LND users.

Please DO NOT REBASE. Instead `git merge` and fix up the merge conflict manually.

PR #115, opened 2022-07-31
bsaes build fix and .gitignore cleanup

tiero:
- cleanup: remove old plugin name
- build fix: lnd-0.14 branch refers to dead git server, point at mirror on github until we don't need it anymore
- go mod tidy

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -33,7 +33,6 @@ cloud.google.com/go/storage v1.6.0/go.mod h1:N7U0C8pVQ/+NIKOBQyamJIeKQKkZ+mxpohl
 cloud.google.com/go/storage v1.8.0/go.mod h1:Wv1Oy7z6Yz3DshWRJFhqM/UCfaWIRTdp0RXyy7KQOVs=
 cloud.google.com/go/storage v1.10.0/go.mod h1:FLPqc6j+Ki4BU591ie1oL6qBQGu2Bl/tZ9ullr3+Kg0=
 dmitri.shuralyov.com/gpu/mtl v0.0.0-20190408044501-666a987793e9/go.mod h1:H6x//7gZCb22OMCxBHrMx7a5I7Hp++hsVxbQ4BYO7hU=
-git.schwanenlied.me/yawning/bsaes.git v0.0.0-20180720073208-c0276d75487e/go.mod h1:BWqTsj8PgcPriQJGl7el20J/7TuT1d/hSyFDXMEpoEo=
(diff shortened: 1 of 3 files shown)


go.sum
```diff
@@ -43,7 +42,6 @@ github.com/NebulousLabs/fastrand v0.0.0-20181203155948-6fb6489aac4e/go.mod h1:Bd
 github.com/NebulousLabs/go-upnp v0.0.0-20180202185039-29b680b06c82/go.mod h1:GbuBk21JqF+driLX3XtJYNZjGa45YDoa9IqCTzNSfEc=
 github.com/OneOfOne/xxhash v1.2.2/go.mod h1:HSdplMjZKSmBqAxg5vPj2TmRDmfkzw+cTzAElWljhcU=
 github.com/Yawning/aez v0.0.0-20180114000226-4dad034d9db2/go.mod h1:9pIqrY6SXNL8vjRQE5Hd/OL5GyK/9MrGUWs87z/eFfk=
-github.com/Yawning/bsaes v0.0.0-20180720073208-c0276d75487e h1:n88VxLC80RPVHbFG/kq7ItMizCVRPCyLj63UMqxLkOw=
 github.com/Yawning/bsaes v0.0.0-20180720073208-c0276d75487e/go.mod h1:3JAJz+vEO82SkYEkAa2lRPkQC7lslUY24HX3929i2Ec=
```
wtogami, line 45: This `h1:3JAJz+vEO82SkYEkAa2lRPkQC7lslUY24HX3929i2Ec=` sum is identical to what can be found in lnd's `v0-14-3-branch` but I'm wondering why it is different from the sum `h1:BWqTsj8PgcPriQJGl7el20J/7TuT1d/hSyFDXMEpoEo=` being removed from the `git.schwanenlied.me` line above. Shouldn't that be an identical sum given they are of the same git hash? I only want to be sure the replacement here is in fact identical. This different sum suggests it isn't?

nepet: I don't get it either. The version ending with `/go.mod` indicates, that the hash is calculated over the `go.mod` file of the module, which they don't have at all. Reference: https://go.dev/ref/mod#go-sum-files

Could it be, that the golang hashdb returns different hashes because the module has no `go.mod` file?

A quick comparison of the md5sum of my locally cached `git.schwanenlied.me` and `github.com/Yawning` proves them to be identical.

nepet (approved):
As the md5sums of the repositories are identical:
Ack d3f02dd4e1855ebec964fce484ed23f73c5c0d06

Issue #110, opened 2022-07-19
Amount of the fee invoice


nepet:
An initiator (__alice__) of a swap should always pay for the on-chain fees. On a swap-out the non-initiating party of the swap (__bob__)  is in charge of the on-chain part. That is why we have the __fee invoice__ as a part of the protocol (see https://github.com/ElementsProject/peerswap/blob/e563b002ed3f66fbce71f8e07037784dfa940716/docs/peer-protocol.md#the-swap_out_agreement-message).

Currently our implementation only addresses the estimated fee for the `opening-tx` and uses this estimate as the amount of the fee invoice. I see two problems here:
1. What about a refund of the on-chain funds, the `csv path`. Even if we can not know the fee that will be used for this tx in the future, we should at least try to estimate it in such a way that the non-initiating swap partner gets some compensation for this case.
2. We do not know the absolute size of the transaction without reserving the inputs. Reserving the inputs upfront would be annoying as we also have to take care that we release them again in the case that the fee invoice is not payed. Also, this way an evil peer could lock up parts of my wallet for the time that we wait for the fee invoice to be payed for.

What do we do about this? An easy first solution could be that we use the following formula for the fee invoice amount:
```
(estimated_fee_opening_tx + estimated_fee_claim_by_csv_tx)*x
```
with some float `x` that we would hard code in a first version but that could be a config and part of the swap negotiation in the future.

In the future we could call this the `premium` with a negotiated factor `x` ^^.

What do you think?

nepet:
As for now: The fee invoice should compensate for the expected fees of the opening tx. This highly depends on the amount of inputs as well as on the type of input.

The general overhead of the transaction is:

| type | size [vbyte] |
| --- | --- |
| version | 4 |
| len in | 1 |
| len out | 1 |
| nLocktime | 4 |
| --- | --- |
| total | 10 |

The outputs of the opening tx have a size of:

| type | size [vbyte] |
| --- | --- |
| Return (P2WPKH) | 22 |
| Script (P2WSH) | 34 |
| Overhead (val(8), len(1)) | 2*9 |
| --- | --- |
| total | 74 

which totals to:
__overhead__ + __outputs__ = __84 vbyte__

If we expect _P2WPKH_ to be the most common input type we add _68 vByte_ for each input. This leads to the following linear expression:

size_P2WPKH = (68*n_inputs+84) vbyte

[...]

PR #126, opened 2022-09-01
Change estimated size for btc opening tx

nepet:
See #110 
This changes the estimated size of the btc opening tx to `350 vByte`

### DIFF
--- a/clightning/clightning_wallet.go
+++ b/clightning/clightning_wallet.go
@@ -213,10 +213,12 @@ func (cl *ClightningClient) GetOnchainBalance() (uint64, error) {
 	return totalBalance, nil
 }
 
-// GetFlatSwapOutFee returns a fee that is the size of an opening transaction
-// with 2 inputs and 2 outputs (p2wsh, p2wpkg change): 218 bytes
+// GetFlatSwapOutFee returns an estimated size for the opening transaction. This
+// can be used to calculate the amount of the fee invoice and should cover most
+// but not all cases. For an explanation of the estimation see comments of the
+// onchain.EstimatedOpeningTxSize.
 func (cl *ClightningClient) GetFlatSwapOutFee() (uint64, error) {
-	return cl.bitcoinChain.GetFee(218)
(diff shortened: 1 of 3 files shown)


yuyaogawa:
I asked a question related to this issue in the discord group, but let me ask this again. `BitcoinFeeTargetBlocks` is 6 to estimate an opening tx , but `TargetConf` is set to 3 to create the actual opening transaction. Wouldn't `BitcoinFeeTargetBlocks` is 3 be better even if this estimated size is changed to `350 vByte`?

wtogami:
> I asked a question related to this issue in the discord group, but let me ask this again. `BitcoinFeeTargetBlocks` is 6 to estimate an opening tx , but `TargetConf` is set to 3 to create the actual opening transaction. Wouldn't `BitcoinFeeTargetBlocks` is 3 be better even if this estimated size is changed to `350 vByte`?

The target number of blocks for fee estimator has a different purpose from the number of confirmations to wait so there is no reason why they must necessarily be the same value.

Fee estimator target is how long you're willing to wait. There is sometimes a large difference in estimated cost between 3 and 6. It might be better for this to be configurable but for now we want to avoid added complications from config variations.

Number of confirmations to wait is only after the tx is in a block. Perhaps this should also be configurable per peer because you may trust your self or your friend while you might demand more time from unknown peers.

Configurability is topics we should discuss for the next protocol alongside the premium model.

yuyaogawa:
> The target number of blocks for fee estimator has a different purpose from the number of confirmations to wait so there is no reason why they must necessarily be the same value.

Yes, I understand this. Please correct me if I am wrong, but...
These variables `BitcoinFeeTargetBlocks` and `TargetConf` are used for fee estimator target to wait BEFORE confirmation. That is why I am asking if these should be the same value. `BitcoinMinConfs` is used for number of confirmations to wait is only AFTER the tx is in a block.

wtogami:
Please pardon me let's bring this discussion back to Discord. It is off topic in the context of reviewing this PR.

grubles:
Tested both LND and CLN BTC swap-outs and can report no issues so far.

PR #130, opened 2022-09-14
Add timestamp in panic log file on startup

### LINKED ISSUE #75: Add date + timestamp to peerswap-panic-log filename
peerswap-panic-log is currently a little confusing because it appends unrelated tracebacks back to back. We should do the same thing as CLN where the datetimestamp is at the end of each file.

kallerosenbaum:
Fixes #75


wtogami:
waiting for @nepet nepet regarding the unrelated reconnect test failure before merge

PR #133, opened 2022-09-16
Make clear ther plugin must go in lightningd datadir

tiero:


grubles:
This doesn't seem entirely correct. You can have the plugins anywhere and specify in the CLN config file with `plugin=/path/to/plugin`. They don't necessarily need to be in the CLN data directory, unless I'm missing something here.

tiero:
> This doesn't seem entirely correct. You can have the plugins anywhere and specify in the CLN config file with `plugin=/path/to/plugin`. They don't necessarily need to be in the CLN data directory, unless I'm missing something here.

I think you right, I tested again, problem is with the official CLN docker image: all the data mounted to it (ie. plugins) will be then owned by root, therefore CLN then fails to use anything outside ~./lightning/plugins.

We can close this, but maybe add a suggestion to put it there for docker users?

wtogami:
BTW are you sure `~` works in the config file? Other path expansion like $HOME doesn't work in there so I would be surprised if `~` works? In my experience I had to include the absolute path to make this work. 

> problem is with the official CLN docker image: all the data mounted to it (ie. plugins) will be then owned by root, therefore CLN then fails to use anything outside ~./lightning/plugins.

Please explain more what do you mean?

The plugin binary itself can be at any path. CLN doesn't need write access to the binary, only read+execute bits.

PeerSwap writes its own data directory within CLN's datadir. Default is `~/.lightning/bitcoin/peerswap`. But then again I don't think path expansion works at all in CLN's config file so I think you need an absolute path if you want anything different from this default.

wtogami:
It isn't clear what the above issue was but I think there is agreement this isn't a general solution. Closing. Please bring discussion to Discord.

PR #145, opened 2022-10-14
Lnd style channel

nepet:
This PR fixes some blocking issues for the UI implementors. The glightning library could not handle 0 values and RTL requested a lnd style channel id in the `listswaps` method return.


wtogami:
ACK

PR #147, opened 2022-10-23
Adjust liquid flat swapout fee

nepet:
We adjust the estimated swap-out tx size that is used for fee calculation to 3000 vBytes. The size is based on 3 inputs and 3 outputs of which 2 are blinded (2672 vBytes), together with a security margin.

### DIFF
--- a/cmd/peerswaplnd/peerswapd/main.go
+++ b/cmd/peerswaplnd/peerswapd/main.go
@@ -16,6 +16,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/elementsproject/peerswap/isdev"
 	"github.com/elementsproject/peerswap/lnd"
 	"github.com/elementsproject/peerswap/log"
 	"github.com/elementsproject/peerswap/version"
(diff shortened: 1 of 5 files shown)


cmd/peerswaplnd/peerswapd/main.go
```diff
@@ -103,6 +104,9 @@ func run() error {
 	}
 	log.Infof("PeerSwap LND starting up with commit %s and cfg: %s", GitCommit, cfg)
 	log.Infof("DB version: %s, Protocol version: %d", version.GetCurrentVersion(), swap.PEERSWAP_PROTOCOL_VERSION)
+	if isdev.IsDev() {
+		log.Infof("Devmode eneabled.")
```
wtogami, line 108: spelling

wtogami (approved):
Looks fine but fix the spelling before merge

grubles:
Tested L-BTC swap-outs from both LND v0.15.3-beta and CLN (e00857827fd5b45a3022753edf855df4765a4e93) signet nodes and haven't run into any issues.

cmd/peerswaplnd/peerswapd/main.go
nepet: 12a372e

PR #153, opened 2022-11-08
cln plugin config file

nepet:
Core lightning needs to be restarted in order to update the peerswap configuration. This can be of a problem if a node has lots of channels and needs some significant amount of time to start up.
This PR introduces a alternate config file that can be used to configure peerswap. This way it is enough to only restart peerswap instead of the whole core lightning node.


nepet:
A `listconfig` command is still missing, will add this soon.

nepet:
Added command `peerswap-listconfig`

wtogami:
I need your opinions. I think it is too confusing to have two separate ways of configuring this. Meanwhile CLN won't fix the underlying problem soon.

How do people feel about removing the CLN config method entirely and instead use the config file?

Does it have a default path to look for its own config file? Same as the policy.conf path?

Being able to specify a different config file path from CLN config seems like a good idea except it has the same problem of confusing behavior where it behaves differently between CLN startup and plugin dynamic start. We can eliminate the confusion entirely by having it look for its own config file at a default path relative to HOME. Is this a good plan though?

grubles:
Hmm I think having a separate config file might be confusing at first only because every other CLN plugin that has configurable options uses the CLN config file already and that is what people are accustomed to.

I haven't come across a plugin that has its own config file so I think it's a question of whether it's worth it to break away from that convention. IMO it should be fine to do so given CLN's userbase is relatively small and we already kind of break that convention with the `policy.conf` file. The best path might be to have both methods of configuring until there is a concrete timeline for CLN being able to reload its config while running. 

That way people can have a separate file if they plan on modifying their config + reloading frequently.

wtogami:
Biggest problem is `peerswap-config-path` *can't work* due to the same reason we wanted a separate config file to begin with. https://github.com/ElementsProject/lightning/issues/5294 CLN behaves differently during startup than if the plugin is started later. CLN is currently incapable of passing config options to plugins during dynamic plugin start.

`clightning_config.go` contains ...
```
        if policyPath == "" {
                wd, err := os.Getwd()
                if err != nil {
                        return nil, err
                }
                policyPath = filepath.Join(wd, "peerswap", "policy.conf")
        }

```

Sadly I think our only choice now is to entirely drop getting options from CLN. It would be the least confusing if peerswap-plugin similarly detected the path from working directory.


- [ ] CLN chooses the working directory for the plugin process so we don't need to hard-code default paths ourselves. (Fact check?)
- [ ] It should log during peerswap initialization like `Using peerswap datadir: /home/btc/.lightning/bitcoin/peerswap`.
- [ ] Note: Blank or missing peerswap.conf is valid. It is possible to use CLN peerswap without config at all if everything including the bitcoind/elementsd .cookie files are at default paths.
- [ ] Let's also consider removing the option for policypath? The plugin needs write access to the datadir due to the `policy.conf` and `swaps` database. Let's instead use the working directory handed to us by CLN and read `peerswap.conf` in that same directory if it exists. There's no good reason to support arbitrary other pathnames?

nepet:
@wtogami You suggested if no config is given (blank or no file) peerswap should try to connect to bitcoind AND elementsd with the default .cookie, if found at the default location.
Up until now the behavior was: If no bitcoin rpc config is given, we connect to the bitcoind that core-lightning is using, therefore we retrieve the bitcoin config from core-lightning.
For elementsd we also did not use the .cookie file but rather just tried the default port on localhost, no user, no password (that should fail). That way operators have to actively enable liquid instead of peerswap trying to connect to liquid without an operator approving.
Any thoughts?

wtogami:
> For elementsd we also did not use the .cookie file but rather just tried the default port on localhost, no user, no password (that should fail). That way operators have to actively enable liquid instead of peerswap trying to connect to liquid without an operator approving.

I have never provided any PeerSwap config options via the CLN config. It currently uses elementsd automatically. Please keep it that way.

wtogami:
`cmd/peerswap-plugin/main.go`
image

You're correct it didn't read the cookie for bitcoind but it did for elementsd. Please keep this current behavior.

nepet:
You are correct! Actually this is still part of the code (`getElementsClient` in main) and working. I'll move this to the config parser to make it more explicit what is going on.

nepet:
I added some tests for cookie authentication.
This should be good to go now.

wtogami:
@grubles please test

Don't merge yet until the CLI syntax changes are also ready to merge. That way we break things only once. I'll handle the documentation and release.

nepet:
@wtogami see #162

grubles:
Running with a blank `peerswap.conf` file does not work. It appears to pull the config from CLN but if CLN uses a cookie file for `bitcoind`, then it by default tries port `0` which is incorrect:

```
plugin-peerswap-plugin: Starting bitcoin client with chain:signet, rpcuser:__cookie__, rpcpassword:be6c9bdcfe376eefc2464447c64dce19d2ee1655fdde66c1cd4bca53b7732802, rpchost:http://127.0.0.1, rpcport:0                   
2023-01-23T17:11:58.573Z INFO    plugin-peerswap-plugin: Exited with error: Post \"http://127.0.0.1:0\": dial tcp 127.0.0.1:0: connect: connection refused
```

The minimum config file for getting both BTC and L-BTC swaps working was this:

```
[Bitcoin]
rpcuser=xxx
rpcpassword=xxx
rpcport=xxx

[Liquid]
rpcport=xxx
rpcwallet=xxx
rpcport=18884
enabled=true
```

nepet:
@grubles I will have a look at this.

wtogami:
* Please rebase.
* Please also add code that exits with a warning if it detects old CLN PeerSwap config options. We don't want both to coexist because that will confuse people.

```UPGRADE NOTICE: Please remove all peerswap* options from CLN config. PeerSwap CLN plugin options are now in peerswap.conf. See documentation for details.
Exiting...
```

wtogami:
peerswap-plugin currently works with zero `peerswap*` options in CLN config. It should behave identically after this change with a blank or missing peerswap.conf where it tries defaults for that chain. CLN does tell plugins what chain?

nepet:
@wtogami Yes, we get this information from the `listconfigs` command, the `bcli` plugin holds a field `network`.

nepet:
> [ ] Let's also consider removing the option for policypath? The plugin needs write access to the datadir due to the `policy.conf` and `swaps` database. Let's instead use the working directory handed to us by CLN and read `peerswap.conf` in that same directory if it exists. There's no good reason to support arbitrary other pathnames?

I don't see why we should not allow for a different location, but I am happy to remove the option If you think otherwise. If we remove this option, how about the db path?

nepet:
Sorry for the messy PR. I got really annoyed by the way we handled the config before and reworked it into some kind of pipes and filter pattern which hopefully will be easier to maintain and to find bugs.

@wtogami @grubles Does auto load from the elementsd cookie file now work for you?

grubles:
With a missing CLN `peerswap.conf`, only Bitcoin swaps are enabled. With a bare minimum LND `peerswap.conf` (with cert + macaroon paths), only Bitcoin swaps are enabled as well.

grubles:
Oh is it because the default is to try port `7041`?

```
plugin-peerswap-plugin: Starting with config: {\"DataDir\":\"/home/user/.lightning/signet\",\"DbPath\":\"/home/user/.lightnin
g/signet/peerswap\",\"PolicyPath\":\"/home/user/.lightning/signet/policy.conf\",\"Bitcoin\":{\"RpcUser\":\"__cookie__\",\"RpcPassword\":\"7184b487cf71af0d053b
bb0d97527588bb7744a32fbcf5a96fbd05ea6a49504d\",\"RpcPasswordFile\":\"/home/user/.bitcoin/signet/.cookie\",\"RpcHost\":\"http://127.0.0.1\",\"RpcPort\":38332,\
"Network\":\"signet\",\"DataDir\":\"/home/user/.bitcoin\"},\"Liquid\":{\"RpcUser\":\"\",\"RpcPassword\":\"\",\"RpcPasswordFile\":\"/home/user/.elements/testne
t/.cookie\",\"RpcHost\":\"http://127.0.0.1\",\"RpcPort\":7041,\"RpcWallet\":\"peerswap\",\"Network\":\"testnet\",\"DataDir\":\"/home/user/.elements\",\"Enable
d\":false}}                                           
```

nepet:
Yes @grubles the „liquid“ default port is assumed. If I remember correctly, the elements and bitcoin signet/testnet/regtest default ports are the same.

grubles:
Do you mean to say that the ports do not change between the different networks? Each has its own specific default port.

nepet:
> Do you mean to say that the ports do not change between the different networks? Each has its own specific default port.

What I meant is, that elementsd and bitcoind both have the same default ports for the different networks (each).

Shall we assume `liquidtestnet` with `18884` or `liquid` with `7041` as default chain if an empty config is given (elementsd defaults to `7041` according to `elementsd --help`)?

grubles:
Based on my log above, it appears like peerswap is able to detect I'm running on Liquid testnet but seems to be using the wrong directory for the `.cookie` file (`/home/.elements/testnet`, should be `/home/.elements/liquidtestnet`). 

If we're able to detect the Elements chain, can we set the appropriate port automatically?

nepet:
> 



> Based on my log above, it appears like peerswap is able to detect I'm running on Liquid testnet but seems to be using the wrong directory for the `.cookie` file (`/home/.elements/testnet`, should be `/home/.elements/liquidtestnet`).
> 
> If we're able to detect the Elements chain, can we set the appropriate port automatically?

Sure! Is there a default port for `liquidregtest` that I can assume? Or do we just fallback to `18884` if the chain is not liquid?
Dir detection is not a problem either I would expand to differentiate between `liquid` `liquidregtest` and `liquidtestnet`.

grubles:
Let's use these ports: 

liquidv1: 7041
liquidtestnet: 18891 (from https://liquidtestnet.com/) 
regtest: 18443 (from elementsd --help)

grubles:
Now with a minimal `peerswap.conf` with:

```
[Liquid]
enabled=true
```
...both BTC and L-BTC swaps are enabled for CLN.
```
2023-02-17T16:23:44.031Z INFO    plugin-peerswap-plugin: Cln synced, continue...
2023-02-17T16:23:44.031Z INFO    plugin-peerswap-plugin: Starting elements client with rpcuser: __cookie__, rpcpassword: 54117b81d494db02941d45adb9d84fa147172ece5259974b5818e2a8ed84c184, rpccookie: /home/user/.elements/liquidtestnet/.cookie, rpcport: 18891, rpchost: http://127.0.0.1
2023-02-17T16:23:44.053Z INFO    plugin-peerswap-plugin: Liquid swaps enabled
2023-02-17T16:23:44.054Z INFO    plugin-peerswap-plugin: Starting bitcoin client with chain:signet, rpcuser:__cookie__, rpcpassword:7184b487cf71af0d053bbb0d97527588bb7744a32fbcf5a96fbd05ea6a49504d, rpchost:http://127.0.0.1, rpcport:38332
2023-02-17T16:23:44.054Z INFO    plugin-peerswap-plugin: Bitcoin swaps enabled
2023-02-17T16:23:44.054Z INFO    plugin-peerswap-plugin: Using gbitcoind estimator
2023-02-17T16:23:44.055Z INFO    plugin-peerswap-plugin: using policy:
2023-02-17T16:23:44.055Z INFO    plugin-peerswap-plugin: allow_new_swaps: true
2023-02-17T16:23:44.055Z INFO    plugin-peerswap-plugin: min_swap_amount_msat: 100000000
2023-02-17T16:23:44.055Z INFO    plugin-peerswap-plugin: reserve_onchain_msat: 0
2023-02-17T16:23:44.055Z INFO    plugin-peerswap-plugin: allowlisted_peers: [035ca2fe4793a5e789ce846062eb4834f573c060d9200ce77544a29b48a0aa5923]
2023-02-17T16:23:44.055Z INFO    plugin-peerswap-plugin: accept_all_peers: false
2023-02-17T16:23:44.055Z INFO    plugin-peerswap-plugin: suspicious_peers: []
2023-02-17T16:23:44.055Z INFO    plugin-peerswap-plugin: 
2023-02-17T16:23:44.121Z DEBUG   035ca2fe4793a5e789ce846062eb4834f573c060d9200ce77544a29b48a0aa5923-connectd: peer_out INVALID 42085
2023-02-17T16:23:44.126Z INFO    plugin-peerswap-plugin: peerswap initialized
```

grubles:
The new `listconfig` command works:

```
$ lightning-cli --signet peerswap-listconfig
{
   "dbpath": "/home/user/.lightning/signet/peerswap",
   "bitcoin.rpcuser": "__cookie__",
   "bitcoin.rpcpassword": "7184b487cf71af0d053bbb0d97527588bb7744a32fbcf5a96fbd05ea6a49504d",
   "bitcoin.rpcpasswordfile": "/home/user/.bitcoin/signet/.cookie",
   "bitcoin.rpchost": "http://127.0.0.1",
   "bitcoin.rpcport": 38332,
   "bitcoin.rpccookiefilepath": "/home/user/.bitcoin/signet/.cookie",
   "liquid.rpcuser": "__cookie__",
   "liquid.rpcpassword": "54117b81d494db02941d45adb9d84fa147172ece5259974b5818e2a8ed84c184",
   "liquid.rpcpasswordfile": "/home/user/.elements/liquidtestnet/.cookie",
   "liquid.rpchost": "http://127.0.0.1",
   "liquid.rpcport": 18891,
   "liquid.rpcwallet": "peerswap",
   "liquid.enabled": true,
   "policypath": "/home/user/.lightning/signet/policy.conf",
   "ConfigFilePath": "/home/user/.lightning/signet/peerswap.conf"
}
```

grubles:
Is the intended default behavior such that the user needs to explicitly enable Liquid swaps or is that supposed to be automatically enabled if Peerswap detects a running `elementsd`? In the PR's current state, I have to create a `peerswap.conf` file with:

```
[Liquid]
enabled=true
```

to get L-BTC swaps working.

nepet:
We always had the `enabled` flag so I assumed that it has to be true to activate l-btc swaps, while we have no flag for btc swaps to enable/disable them. If l-btc should be enabled by default if an elementsd installation is detected we would have to change to a `disabled` flag as the default boolean value if non is given is `false`.

wtogami:
The current code works with L-BTC swaps enabled with no config options at all. If it detects elementsd it uses it. Please keep it that way.

nepet:
Okay, then I don't see the reason for an `enabled` flag altogether. Shall we change it to a `disabled` flag to provide a way to disable L-BTC swaps?

wtogami:
Yes

nepet:
Rebased to include the ci fix.

nepet:
Can add that if you like, currently it only logs if the `disabled` option is set.

grubles:
```
2023-03-15T15:40:08.912Z INFO    plugin-peerswap-plugin: Cln synced, continue...
2023-03-15T15:40:08.912Z INFO    plugin-peerswap-plugin: Liquid swpas disabled
2023-03-15T15:40:08.912Z INFO    plugin-peerswap-plugin: Starting bitcoin client with chain:signet, rpcuser:__cookie__, rpcpassword:401d998cd06fae00bcdfeabea405ca2bc84dac3fa0c55783a15b438a61a5040b, rpchost:http://127.0.0.1, rpcport:38332
```

TY. Message is back but there's a typo "swpas".

wtogami:
My mainnet deployment has `bitcoind`, `lightningd`, and `elementsd` running with default pathnames within the same non-root user.

bitcoind: `~/.bitcoin/`
lightnind: `~/.lightning/`
elementsd: `~/.elements/`

CLN's ~/.lightning/config contains zero `peerswap-*` options. I expect peerswap after this PR to behave the same with auto-detected defaults from the environment but it doesn't.

```
2023-03-17T19:48:06.830Z INFO    plugin-peerswap: PeerSwap starting up with commit 9b4e55b04d804fb6a393bd9cf1504c3e831ab3d1
2023-03-17T19:48:06.830Z INFO    plugin-peerswap: DB version: v0.2, Protocol version: 3
2023-03-17T19:48:06.830Z INFO    plugin-peerswap: Using data dir: /home/bitcoin/.lightning/bitcoin
2023-03-17T19:48:06.853Z INFO    plugin-peerswap: Could not read config: can not get network dir for bitcoin network: bitcoin
2023-03-17T19:48:06.853Z INFO    plugin-peerswap: Exited with error: can not get network dir for bitcoin network: bitcoin
```

wtogami:
We never had anything in our datadir named `mainnet`.

I did a search and replace `/s/mainnet/bitcoin/g` within `clightning/config.go` and it got a little further.

```
2023-03-17T20:43:39.812Z INFO    plugin-peerswap: PeerSwap starting up with commit 9b4e55b04d804fb6a393bd9cf1504c3e831ab3d1
2023-03-17T20:43:39.812Z INFO    plugin-peerswap: DB version: v0.2, Protocol version: 3
2023-03-17T20:43:39.812Z INFO    plugin-peerswap: Using data dir: /home/bitcoin/.lightning/bitcoin
2023-03-17T20:43:39.847Z INFO    plugin-peerswap: Could not read from bitcoin cookie: /home/btc/.bitcoin/bitcoin/.cookie
2023-03-17T20:43:39.877Z INFO    plugin-peerswap: Using core-lightning version v23.02.2
2023-03-17T20:43:39.877Z INFO    plugin-peerswap: Waiting for cln to be synced...
2023-03-17T20:43:49.923Z INFO    plugin-peerswap: Cln synced, continue...
2023-03-17T20:43:49.923Z INFO    plugin-peerswap: Starting elements client with rpcuser: __cookie__, rpcpassword: REDACTED, rpccookie: /home/bitcoin/.elements/liquidv1/.cookie, rpcport: 7041, rpchost: http://127.0.0.1
2023-03-17T20:43:50.013Z INFO    plugin-peerswap: Liquid swaps enabled
2023-03-17T20:43:50.014Z INFO    plugin-peerswap: Exited with error: unknown bitcoin network
```

`/home/bitcoin/.bitcoin/bitcoin/.cookie` isn't the default path where the `.cookie` file is written. By default it is created at `~/.bitcoin/.cookie`

wtogami:
As an ugly hack I got it to start by renaming `s/mainnet/bitcoin/` in `clightning/config.go` and that one spot in `cmd/peerswap-plugin/main.go`. Then I removed `netdir` from the pathname before `.cookie`.

But this isn't the solution because the peerswap spec calls it `mainnet`.

I then discovered a new problem. Previous peerswap would put its database and policy.conf in `~/.lightning/bitcoin/peerswap/` but now it's putting it in `~/.lightning/bitcoin/`. Not 100% sure if that happened as the result of my temporary hacks. Please check.

Are we currently hard coding the netdir based upon the network name? CLN provides plugins the directory to use for writing as the `cwd` or current working directory. We shouldn't second guess that. This way peerswap would work automatically with any non-default specified `--lightning-dir=`. 

Tasks
== 
- [ ] Special case pathname "bitcoin" while keeping peerswap's network name as "mainnet".
- [ ] Investigate where bitcoind puts the .cookie file for mainnet, it may differ with $HOME default vs specified `-datadir`. Default is `~/.bitcoin/.cookie`.
- [ ] Ensure peerswap/ directory is used within the CLN provided working directory.
- [ ] Get rid of `peerswap-db-path`. Is there a good reason to use it?

wtogami:
We have a possible alternative of asking CLN how it is accessing bitcoind. However maybe we don't want to use this because peerswap isn't guaranteed to be using `bcli`. There exist a few drop-in replacements including block explorer lookup.

`lightning-cli listconfigs`

```
      {
         "path": "/home/bitcoin/libexec/c-lightning/plugins/bcli",
         "name": "bcli",
         "options": {
            "bitcoin-datadir": null,
            "bitcoin-cli": null,
            "bitcoin-rpcuser": null,
            "bitcoin-rpcpassword": null,
            "bitcoin-rpcconnect": null,
            "bitcoin-rpcport": null,
            "bitcoin-retry-timeout": null,
            "commit-fee": null
         }
      },
```

grubles:
Something weird is happening with using `elementsd` via RPC too:

```
2023-03-17T21:44:19.977Z INFO    plugin-peerswap: Could not read config: can not get liquid network dir for bitcoin network: bitcoin
2023-03-17T21:44:19.977Z INFO    plugin-peerswap: Exited with error: can not get liquid network dir for bitcoin network: bitcoin
```

nepet:
@wtogami Thanks for testing and reporting!
1. Indeed,  core-lightning reports the network `bitcoin` when using `main/mainnet` chain that messed up with the path finding and Is the reason for your first error and also @grubles latest report.
2. I can also confirm the default path to the cookie file is `/$HOME/.bitcoin/.cookie` for empty config. I guess this is what we should support for the default case then. If people are using `--datadir` the cookie file is inside of the datadir (with optional network subdir), e.g. someone using `-datadir="$HOME/.bitcoin/mydir` results in the cookie file path `$HOME/.bitcoin/mydir/<optional-network>/.cookie`. So IF we get the datadir e.g. by core-lightnings `listconfigs` command, we can assume the cookie file at this place. Otherwise the user needs to set peerswap config to point to the location of the cookie file. The cookie file gets wiped on shut down of bitcoind so we don't have to care about old state.
3. Elements default dir (empty config) is `$HOME/.elements/liquidv1` which is where the cookie file is located.
4. About the peerswap database file location: We get the peerswap datadir from the `cwd` that core-lightning sets for the peerswap child process. That is the `lightning-dir`. I was in believe that we said we want a flat directory structure and do not want to create a  `lightning-dir/peerswap` directory anymore to store peerswap related data as the `peerswap.conf` was located next to lightning relevant files inside of the `lightning-dir`. We could however place all peerswap related data (db, policy.conf) in a peerswap subdir.

nepet:
I agree that we can get rid of the `peerswap-db-path`. People can just move their db file if the assumed location changes instead of setting an option pointing to the file.

nepet:
@wtogami About the `peerswap` dir inside of the `lightning-dir. Shall this dir contain all peerswap related files? That would be:
- config `peerswap.conf`
- policy `policy.conf`
- db `peerswap`

Correct?

wtogami:
Yes

wtogami:
> [Add PeerswapDir to config](https://github.com/ElementsProject/peerswap/pull/153/commits/9f8cc6f4452a9ce4f96bc944f5886adfe4031193)

When I suggested removing the peerswap datadir option it was because I suggested it always create a "peerswap" directory in the current working directory handed it to the plugin by CLN, as was the default behavior prior to this PR.

Why make this configurable? I don't feel strongly about this but if there isn't a good reason I'd prefer we don't have an option.

wtogami:
> I was in believe that we said we want a flat directory structure and do not want to create a `lightning-dir/peerswap` directory anymore

I don't recall this discussion. It is a valid design choice but now it has a switching cost for existing deployments. Is there a good reason in the long-term to not have all the peerswap config, policy and state in its own child directory?

grubles:
Using a fresh CLN datadir and using CLN v23.02.2 on signet:
	
* Peerswap sub diretory created
* BTC swaps automatically enabled. `bitcoind` `.cookie` file detected. 
* BTC and L-BTC swaps automatically enabled. Both `bitcoind` and `elementsd` `.cookie` files detected when `elementsd` is running.
* All working without a `peerswap.conf` file

grubles:
Although on mainnet, the plugin exits:

`2023-04-10T19:43:47.270Z INFO    plugin-peerswap: Exited with error: unknown bitcoin network`

wtogami:
* please test that blank config file is equal to missing.
* What peerswap calls mainnet internally is different from CLN. We can't rename it in peerswap because it's in the wire protocol. So we need to hatdcode substitution in peerswap.

grubles:
Tested with a blank `peerswap.conf` at:

~/.lightning/signet/peerswap/peerswap.conf

Both BTC and L-BTC swaps were enabled.

grubles:
Testing newest commits against CLN v23.02.2 and LND v0.16.1-beta:

* Peerswap subdirectory is created, swap database and policy file are created there
* Automatically finds `bitcoind` and `elementsd` signet cookie files, enables BTC and L-BTC swaps.
* Blank and missing `peerswap.conf` work
* Peerswap detects Bitcoin mainnet successfully, finds `bitcoind` / `elementsd` cookie files
* Tested signet BTC and L-BTC swaps with both CLN and LND as swap initiators

Issue #140, opened 2022-10-06
Peerswap plugin killed on node startup 


ShahanaFarooqui:
Peerswap plugin on Core Lightning is being killed on startup for the node with below error:

```
2022-10-05T17:16:56.975Z DEBUG   plugin-peerswap-plugin: poll_service: could not send request_poll msg: -32600:Peer is not connected
2022-10-05T17:16:56.975Z DEBUG   0391cd7ae2eca090c5f807d71ab600bea5590796b56b21b44de596f9be42b7fd96-connectd: peer_out INVALID 42085
2022-10-05T17:16:56.979Z INFO    plugin-peerswap-plugin: [Swap:96500ce1c7824710450a325ee0e9894d619ecfe0d666abe8300e8ef53183aea1]: Recovering from state State_SwapOutSender_AwaitAgreement
2022-10-05T17:16:56.980Z DEBUG   plugin-peerswap-plugin: [FSM] event:id: 96500ce1c7824710450a325ee0e9894d619ecfe0d666abe8300e8ef53183aea1, Event_ActionFailed on State_SwapOutSender_AwaitAgreement
2022-10-05T17:16:56.980Z INFO    plugin-peerswap-plugin: plugin quitting, error: event rejected
2022-10-05T17:16:56.981Z INFO    plugin-peerswap-plugin: Killing plugin: exited during normal operation
```

CLN node config:
```
network=signet
log-level=debug
bitcoin-rpcuser=***
bitcoin-rpcpassword=***
plugin=/.../peerswap/peerswap-plugin
peerswap-db-path=/.../l1-signet/swap
peerswap-policy-path=/.../l1-signet/policy.conf
``` 

I am unable to start peerswap on this node anymore.

ShahanaFarooqui:
@nepet 
1: This issue is persistent even without changing networks (as oppose to our discussion in today's meeting). 
2: Steps to reproduce:
- Start CLN with Peerswap plugin
- Execute a successful `swap-out`/`swap-in` from the node.
- Stop CLN while this swap is in active state.
- Restart CLN. 
- peerswapd will crash with error message `Killing plugin: exited during normal operation`.

nepet:
More insight: This seems to be the case when recovering from a  State_SwapInSender_AwaitClaimPayment. On recovery the following nil pointer exception is recorder
```
panic: runtime error: invalid memory address or nil pointer dereference
[...]

PR #155, opened 2022-11-10
Add nil check on gettxout

nepet:
In some rare occasions the gettxout command returns nil, we do not want to panic in these cases.


nepet:
@ShahanaFarooqui could you please check if this solves #140 for you?

ShahanaFarooqui:
@nepet Perfect, it didn't break anymore. Thank you for the fix.

PR #159, opened 2022-12-19
Replace MPP with single-part `sendpay`

nepet:
We replace the mpp mechanic with a simpler and more robust single payment via the desired channel as route.
@grubles Is this helpful for you?


wtogami:
@nepet this would be a good time to also add the startup checks that make wumbo mandatory? It could lead to unexpected for some people without wumbo so it's best to just eliminate that as a confounding factor.

grubles:
Thanks. Will give it some testing!

nepet:
@wtogami  Another idea handling wumbo until the patch is merged: Check if wumbo is enabled and if not: reject any swap with `amt > 2^32 msat` with a message: `need to enable --large-channels for swaps > 4M sats`. This way we would not force the node operateor to enable wumbo if they want to use peerswap.

grubles:
A quick mainnet test works, where previously the BTC swap-out would coop fail due to a MPP timeout. Will continue testing.

wtogami:
> Another idea handling wumbo until the patch is merged: Check if wumbo is enabled and if not: reject any swap with amt > 2^32 msat with a message: need to enable --large-channels for swaps > 4M sats. This way we would not force the node operateor to enable wumbo if they want to use peerswap.

Good idea.

nepet:
We now reject a swap with an amount that exceeds `2^32 msat` if the option `--large-channels` is not enabled. I added tests for the combinations _swap-in/-out -> swap amount above/equals `2^32 msat` -> `--large-channels` enabled/disabled_
This should be good to go now

grubles:
Tested more extensively using signet and a few times on mainnet. The `sendpay` patch appears to work.

The `> 2^32 msat` check also works with CLN as the swap-out initiator, but not as the swap-in initiator. 

I can also swap-out `> 2^32` msats with LND as the initiator, but not as the swap-in initiator. 

This is with neither node having wumbo / large-channels enabled.

wtogami:
The detection being one sided and inconsistent doesn't help the underlying confusion that we aimed to eliminate. Should we ship this PR as is or remove the incomplete check?

grubles:
IMO removing it for now sounds like a good idea since it doesn't completely reject >2^32 msat swaps if wumbo is not enabled.

nepet:
Oh sorry for the confusion, I never figured that this was also meant to block lnd payments as they do not have the same legacy as cln. I can add the same constraint also for lnd, this would not be too much work!

nepet:
Also, the check is always performed on the side that pays the Lightning invoice as it is an somewhat abstract „can spend x sat“ check. I can also check on init of all swaps if the channel partner can spend, but this would then also affect/block lnd payments that would be fine.

grubles:
The `cancel_message` could be clarified to say something like "the CLN peer needs to enable --large-channels". If I'm an LND peer initiating a swap, it will error telling me to enable `--large-channels` which isn't an LND config option. 

example from an LND node:

```
$ ./builds/peerswap/out/pscli swapin --sat_amt 5000000 --channel_id 137426858844160000 --asset lbtc                             
2023/01/09 22:11:16 rpc error: code = Unknown desc = swap amount is 5000000000: need to enable option '--large-channels' to swap amounts larger than 2^32 msat 
```

nepet:
Oh I see, I can definitely give it a better error message.

wtogami:
Merged despite the confusion error message. If both sides have wumbo enabled it isn't a problem. Newer CLN versions will change the behavior in the case where wumbo is not enabled?

grubles:
Opened a new issue for the confusing error message: https://github.com/ElementsProject/peerswap/issues/166

PR #160, opened 2022-12-20
Log errors from "main scope"

nepet:
We had a hard time debugging certain errors as the cln logs only reported `plugin-peerswap-plugin: Killing plugin: exited during normal operation`. The reason for this was that the errors could not be passed to the cln log as the errors escaped the scope that handled the json rpc connection and messaging to core lightning.

This PR adds logs to errors that are thrown in the `run()` routine. Therefore we add a wrapper around `run()` that initializes the plugin and can use the json rpc interface to log an error that is thrown by peerswap.

### DIFF
--- a/cmd/peerswap-plugin/main.go
+++ b/cmd/peerswap-plugin/main.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	log2 "log"
 	"os"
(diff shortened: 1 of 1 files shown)


nepet:
I am going to merge this now even if it is not reviewed as this is super helpful for debugging and does not touch any api or config, but just the way we start up peerswap internally.

PR #162, opened 2023-01-03
Change oreder of parameters in cln swap commands

nepet:
Cln prefers the short channel id at first parameter, we follow this to avoid confusion.

### DIFF
--- a/clightning/clightning_commands.go
+++ b/clightning/clightning_commands.go
@@ -184,8 +184,8 @@ func (s *LiquidSendToAddress) LongDescription() string {
 
 // SwapOut starts a new swapout (paying an Invoice for onchain liquidity)
 type SwapOut struct {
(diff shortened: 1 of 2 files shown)


grubles:
Taking a look at some of the channel-related CLN calls (e.g., `close`, `setchannel`, `fundchannel`, `listchannels`, etc.), they all expect an `id` parameter first, so I think this change would help.

grubles:
I can't see where we can improve upon the other peerswap commands in the context of matching CLN conventions. Does anyone else have any ideas?

nepet:
If there are no more complaints, ideas, I am going to merge this soon.

wtogami:
Go ahead with merge

wtogami:
Merged but there remains the confusing error message

PR #163, opened 2023-01-05
Reduce unconnected peer log spam

nepet:
This log was related to sending out poll messages to our peers. If a peer is not connected there is no need in sending a message to it. We only care about connected peers and stop sending these messages to knowingly unconnected peers to reduce log spam.

### DIFF
--- a/clightning/clightning.go
+++ b/clightning/clightning.go
@@ -497,7 +497,9 @@ func (cl *ClightningClient) GetPeers() []string {
 
 	var peerlist []string
 	for _, peer := range peers {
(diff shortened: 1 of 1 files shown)


grubles:
Does the log spam only occur when starting CLN? I don't seem to see it any other time. 

Tested locally and can confirm there is no `DEBUG   plugin-peerswap-plugin: poll_service: could not send request_poll msg: -32600:Peer is not connected` spam when starting a CLN node and a peerswap peer is not connected.

grubles:
Also checked for logs that occur once a peer disconnects _after_ CLN has started. Looks good.

PR #169, opened 2023-02-06
More descriptive `--large-channel` error message

### LINKED ISSUE #166: swap cancel_message is confusing
The `cancel_message` when the CLN peer does not have `wumbo` / `large-channels` enabled could be clarified to say something like "the CLN peer needs to enable --large-channels". If I'm an LND peer initiating a swap, it will error with a message that could be misinterpreted to mean the LND peer needs to enable `--large-channels`, which isn't an LND config option.

Example from an LND node:

```
$ ./builds/peerswap/out/pscli swapin --sat_amt 5000000 --channel_id 137426858844160000 --asset lbtc                             
2023/01/09 22:11:16 rpc error: code = Unknown desc = swap amount is 5000000000: need to enable option '--large-channels' to swap amounts larger than 2^32 msat 
```

nepet:
The error we return when we receive an swap-in request that we can not accept becaus of a missing `--large-channel` flag was confusing for lnd peers, as they don't know this option.
We are now more specific about this.

resolves #166

### DIFF
--- a/swap/services.go
+++ b/swap/services.go
@@ -47,6 +47,7 @@ type LightningClient interface {
 	AddPaymentNotifier(swapId string, payreq string, invoiceType InvoiceType)
 	RebalancePayment(payreq string, channel string) (preimage string, err error)
 	CanSpend(amountMsat uint64) error
+	Implementation() string
 }
 
 type TxWatcher interface {
--- a/clightning/clightning.go
+++ b/clightning/clightning.go
@@ -180,6 +180,12 @@ func (cl *ClightningClient) CanSpend(amtMsat uint64) error {
 	return nil
 }
 
+// Implementation returns the name of the lightning network client 
+// implementation.
+func (cl *ClightningClient) Implementation() string {
+	return "CLN"
+}
+
 // CheckChannel checks if a channel is eligable for a swap
 func (cl *ClightningClient) CheckChannel(channelId string, amountSat uint64) error {
 	funds, err := cl.glightning.ListFunds()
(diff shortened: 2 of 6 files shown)


swap/services.go
```diff
@@ -47,6 +47,7 @@ type LightningClient interface {
 	AddPaymentNotifier(swapId string, payreq string, invoiceType InvoiceType)
 	RebalancePayment(payreq string, channel string) (preimage string, err error)
 	CanSpend(amountMsat uint64) error
+	Implementation() string
```
wtogami, line 50: I might be missing something where in the code does it actually print Implementation() during runtime?

nepet: https://github.com/ElementsProject/peerswap/blob/23ed52b4997c4e6cd38615699016cc5a7ea25125/swap/service.go#L437-L447

wtogami: So this is reported over the wire to the peer? If so should that be mentioned in the spec too?

grubles:
Looks good to me:

```
$ ./peerswap/out/pscli swapin --sat_amt 10123817 --channel_id 137426858844160000 --asset lbtc 
2023/02/14 15:14:31 rpc error: code = Unknown desc = from the CLN peer: swap amount is 10123817000: need to enable option '--large-channels' to swap amounts larger than 2^32 msat
```

swap/services.go
nepet: Mh, good question. I think that the error message is implementation specific, we have the special case here, that the error message is related to the cln option `--large-channel`. I guess that this is something we should loosely communicate or agree on with other implementations (acinq/eclair to be named) instead of enforcing it by spec. The reason for the informative error is to avoid confusion if the swap partner is a different implementation (lnd currently).

PR #172, opened 2023-03-02
Docker related changes

grubles:
Adding a disclaimer to the README stating we're not responsible for how users deploy Peerswap and adding some supporting documentation for how to deploy using Docker. Also removing `tag.yml` since we already test every commit in master and because we won't build and distribute an official Peerswap Docker image. Removing the `Dockerfile` in the root directory too. `.goreleaser.yml` doesn't look used anywhere, so it's getting purged too. 

(Closed the other PR because I intended to group the changes separately but figured that wasn't necessary)


nepet (approved):
Seems good to me!

Issue #72, opened 2022-06-03
Change `peerswap-plugin` binary name to `peerswapcln`


wtogami:
It was originally named `peerswap` but later renamed to the current `peerswap-plugin`.

```
2022-06-03T00:49:43.585Z INFO    plugin-peerswap-plugin: [Swap:fe679c6cce34905c15e31db98a1242c8dee0d31b66997c18c4a3f84a3371447c] Paid Feeinvoice of 258 sats
```

It looks stupid in logs. I don't want to look at the redundant word in logs for years. We're changing this back before the release of v0.3.0.

My bad.

wtogami:
It was renamed from `peerswap` to `peerswap-plugin` because I thought the name looked ambiguous, one character different from the LND `peerswapd` binary.

How about `peerswapcln` for the plugin binary name? Shorter, not redundant in CLN logs, and not ambiguous as to what it is.

openoms:
Naming is hard, but I agree to change the plugin name to `peerswapcln` - already using that as the config option in raspiblitz https://github.com/rootzoll/raspiblitz/pull/3202.

I think it would make the naming even more consistent if the `peerswapd` binary for LND would be called `peerswaplnd` avoiding the confusion that it might cause if someone thinks it is needed for CLN also. 
I like `pscli` as it is short and is somewhat a self-explanatory as a pair of the `peerswapd` / `peerswaplnd` daemon.

wtogami:
Thank you for expressing opinions!

Yeah I like the idea of renaming the LND binary as well.

PR #173, opened 2023-03-06
Rename peerswap-plugin

grubles:
This changes the name of the `peerswap-plugin` binary to `peerswap` and addresses #72.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,10 +26,10 @@ dist/
 .ci/docker/liquid-config2/**
 
 # build output
-peerswap-plugin
+peerswap
 out/peerswapd
-out/peerswap-plugin
+out/peerswap
 out/pscli
 out/test-builds/peerswapd
-out/test-builds/peerswap-plugin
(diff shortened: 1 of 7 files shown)


nepet:
It seems that the tests are still looking for the `peerswap-plugin` binary.

grubles:
Looks like that fixed it.

grubles:
@nepet This looks like it can be merged as well. Thoughts?

nepet:
Had another look at the whole repo. We need some more renaming:
- `.gitignore` needs to ignore the `peerswap` binary (currently it ignores `peerswap-plugin`
- the `startup_regtest.sh` script in contrib still points to `peerswap-plugin`
- `setup_cln.md` has references to the `peerswap-plugin` binary
- also does `signetguide_cln`
- and also `upgrade.md`

Sorry for the late reply, I hope I found all the missing pieces that need to be renamed this time ^^.

nepet (approved):
Sgtm! :+1:

PR #174, opened 2023-03-06
fix markdown in README and docs/docker.md

grubles:
My local `vim` incorrectly displayed the markdown.


grubles:
@nepet Are the tests manually canceled? I think we can merge this one since it's a minor markdown change.

nepet:
You are correct this pr does not need to be tested. To avoid spending CI time on "doc-only" prs one can add e.g .`[skip ci]` to the commit message https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs.
Go ahead and merge If you like :+1:

grubles:
Ah TIL! Will do that for future docs PRs.

Issue #89, opened 2022-07-06
peerswap-plugin compile error on aarch64


openoms:
Testing on aarch64 Debian (RaspiBlitz v1.8.0rc3)
```
₿ uname -a
Linux raspberrypi 5.15.32-v8+ #1538 SMP PREEMPT aarch64 GNU/Linux

₿ go version
go version go1.17.3 linux/arm64
```
Running:
```
PATH=/usr/local/go/bin/:$PATH
make cln-release
```
```
go build -o peerswap-plugin -ldflags "-X main.GitCommit=e9d5a565e063208bf88ecea0af04aa48e75e15ce" ./cmd/peerswap-plugin/main.go
go: downloading github.com/btcsuite/btcd v0.22.0-beta.0.20211005184431-e3449998be39
go: downloading github.com/elementsproject/glightning v0.0.0-20220502095901-74f998c51367
go: downloading github.com/vulpemventures/go-elements v0.3.7
go: downloading go.etcd.io/bbolt v1.3.6
go: downloading github.com/btcsuite/btcutil v1.0.3-0.20210527170813-e2ba6805a890
go: downloading github.com/btcsuite/btcutil/psbt v1.0.3-0.20210527170813-e2ba6805a890
go: downloading github.com/jessevdk/go-flags v1.5.0
go: downloading golang.org/x/sys v0.0.0-20210915083310-ed5796bab164
go: downloading github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
go: downloading golang.org/x/crypto v0.0.0-20211117183948-ae814b36b871
go: downloading github.com/vulpemventures/go-secp256k1-zkp v1.1.5
go: downloading github.com/vulpemventures/fastsha256 v0.0.0-20160815193821-637e65642941
# command-line-arguments
cmd/peerswap-plugin/main.go:627:8: undefined: syscall.Dup2
make: *** [Makefile:62: cln-release] Error 2
```

It is working on amd64 using the same scripts and software versions.

Edit: 
see the issue in go: https://github.com/golang/go/issues/10235
and a fix in another project: https://github.com/sevlyar/go-daemon/commit/2f19e76436a8afbb39ef64629d543cf07cddc7a5
more: https://stackoverflow.com/questions/55785908/how-to-use-os-or-arch-specific-code-in-go

nepet:
Seems more like `Dup2` is not available which we use to copy the golang panics to a separate log file

wtogami:
[...]

PR #176, opened 2023-03-09
cln: make Dup3 platform independent

JssDWt:
Dup3 is not available on Mac. Use the sys/unix package for a platform independent version.

### DIFF
--- a/cmd/peerswap-plugin/main.go
+++ b/cmd/peerswap-plugin/main.go
@@ -7,12 +7,12 @@ import (
 	glog "log"
 	"os"
 	"path/filepath"
-	"syscall"
 	"time"
 
 	"github.com/elementsproject/peerswap/isdev"
 	"github.com/elementsproject/peerswap/log"
 	"github.com/elementsproject/peerswap/version"
+	"golang.org/x/sys/unix"
 
 	"github.com/vulpemventures/go-elements/network"
 
@@ -488,7 +488,7 @@ func setPanicLogger() (func() error, error) {
 		return nil, err
 	}
 
-	err = syscall.Dup3(int(panicLogFile.Fd()), int(os.Stderr.Fd()), 0)
+	err = unix.Dup2(int(panicLogFile.Fd()), int(os.Stderr.Fd()))
 	if err != nil {
 		return nil, err
 	}


nepet (approved):
I have no idea if this is gona conflict with other unix/linux distributions but it seems to work for me on jammy and OsX 13.2.1.

grubles:
Will this break aarch64? See #89

nepet:
Thanks @grubles, this seems to be more complicated than I thought. We can set OS dependent build constraints https://pkg.go.dev/cmd/go#hdr-Build_constraints to at least support `aarch64` and `x86_64` linux and darwin systems. Does anyone have a better solution to capture panic logs?

nepet (changes requested):
We need to use OS dependent build constraints https://pkg.go.dev/cmd/go#hdr-Build_constraints as `aarch64` systems apperently only support `dup3`. See #89.

grubles:
It actually appears to build fine on my aarch64 Debian 11 machine.

nepet:
@grubles Thanks for testing. It looks like Raspiblitz is using ubuntu 18.04 (based on debian 10) for its base, as far as I can tell from their repository. According to https://github.com/ngaut/log/issues/4 this still is an ongoing issue.
I guess it is best to decide on the OS via build constraints (freebsd is also possible) to decide between dup2/dup3

JssDWt:
The Dup2 in the golang.org/x/sys/unix package appears to contain a platform independent version of Dup2/Dup3. Look at the source [here](https://cs.opensource.google/go/x/sys/+/refs/tags/v0.6.0:unix/syscall_linux.go;l=1817), in the left lane you see there's a separate syscall file for most (all?) platforms there.

nepet (approved):
> The Dup2 in the golang.org/x/sys/unix package I used appears to contain a platform independent version of Dup2/Dup3. Look at the source [here](https://cs.opensource.google/go/x/sys/+/refs/tags/v0.6.0:unix/syscall_linux.go;l=1817), in the left lane you see there's a separate syscall file for most (all?) platforms there. So on linux Dup2 actually calls Dup3 under the hood, but not on darwin, and so forth.

Oh, you are right! The golang.org/x/sys/uinx package is doing already exactly what we proposed. Thanks for pointing this out, so there is no need to change anything and we can merge.

grubles:
ACK.

JssDWt:
rebased

nepet:
Sorry to bother you but could you rebase again? We can merge it now since the ugly config pr was merged.

JssDWt:
@nepet rebased

PR #179, opened 2023-03-16
tests: fix race condition in RemoveUnseen test

JssDWt:
By passing 'now' to the RemoveUnseen function, it becomes a little more stable to test.


nepet (approved):
Ack 8c1652aa58fbbf70818bdd8888135b585fb0efe7

PR #188, opened 2023-05-31
Fix `listpeers` response

nepet:
The response produced an `NaN` error unmarshalling the response. A possible cause could have been the use of deprecated values in a division operation.
This PR changes to the new cln msat values and removes the possible zero division. Everyone can calculate the removed percentage on demand.

### DIFF
--- a/clightning/clightning_commands.go
+++ b/clightning/clightning_commands.go
@@ -655,11 +655,10 @@ func (l *ListPeers) Call() (jrpc2.Result, error) {
 			for _, channel := range peer.Channels {
 				if c, ok := fundingChannels[channel.ShortChannelId]; ok {
 					peerSwapPeerChannels = append(peerSwapPeerChannels, &PeerSwapPeerChannel{
-						ChannelId:       c.ShortChannelId,
-						LocalBalance:    c.ChannelSatoshi,
-						RemoteBalance:   uint64(c.ChannelTotalSatoshi - c.ChannelSatoshi),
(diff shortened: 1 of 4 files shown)


wtogami:
Is this compatible with earlier versions of CLN?

wtogami:
Tested ACK

Remaining question is do we need to bump the minimum CLN version?

nepet:
This fix should also be compatible with `v23.02.x`

grubles:
Tested ACK on v23.05

wtogami:
> compatible with `v23.02.x`
Does PeerSwap detect earlier versions and exit with an informative message?

nepet:
> > compatible with `v23.02.x`
> 
> Does PeerSwap detect earlier versions and exit with an informative message?

min version is set to `v23.02`
https://github.com/nepet/peerswap/blob/f6ed0636d3ad82eed126ea4a4a5f95b092e7cd74/cmd/peerswap-plugin/main.go#L39-L41

If the version is lower, the plugin exits the main routine with:
```
log.Debugf("Core-lighting version %s is not supported, min version is v%s", nodeInfo.Version, minClnVersion)
		return fmt.Errorf("Core-Lightning version %s is incompatible", nodeInfo.Version)
```
Maybe it would be better to log this as `info` instead of `debug`.


When switching to generated bindings at some point in the future, the better place to check for version compatibility would be right after we get the `init` message from cln. If the version is incompatible we should respond with a `"result": {"disabled": "meaningful reason"}` which will disable and kill the plugin before we even started the main loop.

wtogami:
Yes info log please. Add that and merge.

PR #189, opened 2023-06-09
Silence noisy poll log

nepet:
Poll log was too noisy when the peer is disconnected


nepet:
Rebased to include the `listpeers` fix

wtogami:
Tested ACK

PR #190, opened 2023-06-09
Remove passwords from logs

nepet:


wtogami:
Tested ACK, please merge then rebase #189 on master so that becomes easier to test

PR #196, opened 2023-06-14
docs: add elementsd setup + manual usage guide

grubles:


wtogami:
> elements-cli sendtoaddress [address] [amount in decimal form, e.g. 0.1 for 0.01000000 L-BTC]

Huh?

grubles:
There I'm trying to clarify how the amount should be in typical BTC terms and not sats.

PR #198, opened 2023-06-18
Revert "Rework recovery"

wtogami:
Reverts ElementsProject/peerswap#192

```
190d3238 swap/fsm.go (kon           2021-07-13 17:45:44 +0200  19) // ErrFsmConfig is the error returned when the fsm configuartion is invalid
190d3238 swap/fsm.go (kon           2021-07-13 17:45:44 +0200  20) // i.e. the fsm does not contain the next state
e7bda7fa fsm/fsm.go  (kon           2021-06-26 08:45:21 +0200  21) var ErrFsmConfig = errors.New("fsm config invalid")
```

This codepath was triggered on a node running PR #192 attempting to do a lbtc swap-out.

Investigating if reversion is necessary. This might actually be an unrelated race condition!


wtogami:
Reverting due to issue #200. Needs more investigation.

PR #203, opened 2023-07-11
Improve Logs

nepet:
A collection of changes how we log. This includes:
- Don't log redundant custom messages, we just log that we received the same message again
- Don't log "message already exists" this is redundant information to the above
- Remove blank lines when sending redundant messages.
- Add information to the "tx does not have enough confirmations"

Please let me know if there are more blank lines or other improvements that would help you scan the logs.


wtogami (approved):
tested ACK

PR #212, opened 2023-08-01
improve payment notifier log message

grubles:
Example:
```
2023-07-29T02:21:24.113Z DEBUG   plugin-peerswap: [Payment Notifier] Error 903:invoice expired during wait
```

This PR is intended to improve the above error message to include the swap ID and change the log level to INFO.


wtogami:
It's difficult to test this on a real node. Do any of our tests test expiration where you can see this print?

grubles:
Tested on signet by trying @wtogami's idea of starting a swap-in and killing the remote peerswap instance before the opening transaction confirmed.

```
2023-08-02T16:01:53.014Z INFO    plugin-peerswap: [Payment Notifier] Error 903:invoice expired during wait, swap d52feff22b93ed0e996fecf0e8da3813cbc730ece531be7e727d671b3273ad1d
```

PR #213, opened 2023-08-02
Fix Possible Polling Deadlock

nepet:
Fixes a potential mutex deadlock that might be the reason for the store node to disappear from the `peerswap-peerlist`. The potential deadlock was introduced in #189.


wtogami:
tested ACK, seems to fix #185

PR #217, opened 2023-08-11
Add CLN config option to disable BTC swaps, fixup LND swap config

grubles:
This adds a config option for the `[Bitcoin]` section in `peerswap.conf` that disables BTC mainchain swaps so nodes can choose to only swap L-BTC if they desire. The existing behavior of enabling BTC swaps if there are no config options set, or if `peerswap.conf` doesn't exist, is also maintained. 

Example:
```
[Bitcoin]
bitcoinswaps=false
```


grubles:
Added documentation to `setup_cln.md`, clarified the commit message, and squashed the two initial commits.

wtogami:
image

I changed my mind. It looks bad to have the same config option look different in the [Bitcoin] and [Liquid] sections. Please rename Bitcoin's option to simply "enabled" to make it match.

wtogami:
Confirmed this patch does the right thing. Just rename the CLN option and it's ready.

PR #218, opened 2023-08-13
Fix test flake

nepet:
Seems that the "only one active swap per channel" test on lnd is flaky. I was not able to reproduce the failing test on my local machine -> Set a more descriptive error message and check on CI run.


wtogami:
Apparently making it print more fixed the problem?

PR #220, opened 2023-08-19
txwatcher: rework StartBlockchainWatcher() exits

grubles:
This should improve upon #219 such that we exit only if we hit an `Authorization failed: Incorrect user or password` error. For other errors we log once. Also added which daemon the error is coming from into the log print.


wtogami:
> For other errors we log once.
Upon successful later connection INFO log "{bitcoind,elementsd} connection successful" then reset logged to zero?

Issue #224, opened 2023-08-24
plugin crashes after update to Core Lightning v23.08


sat-hoarder:
Plugin will load on initial startup and crash on first lightning-cli call.  Panic-log:

Server started 2023-08-24T03:52:49Z
panic: interface conversion: interface {} is float64, not string

goroutine 1 [running]:
github.com/elementsproject/peerswap/clightning.BitcoinFallbackFromClnConfig.func1(0xc0004b4a00)
        /root/peerswap/clightning/config.go:274 +0x46d
github.com/elementsproject/peerswap/clightning.(*Pipeline).Run(0xc0002a5af0)
        /root/peerswap/clightning/config.go:450 +0xde
github.com/elementsproject/peerswap/clightning.GetConfig(0xc0003fe000)
        /root/peerswap/clightning/config.go:432 +0x69b
main.run({0x103fcd8?, 0xc0001a6eb0}, 0xc0003fe000)
        /root/peerswap/cmd/peerswap-plugin/main.go:130 +0x208
main.outer()
        /root/peerswap/cmd/peerswap-plugin/main.go:101 +0x1e5
main.main()
        /root/peerswap/cmd/peerswap-plugin/main.go:55 +0x137

wtogami:
What version of peerswap? We've been heavily using recent and current master branch. It's close to release candidate.

sat-hoarder:
cloned a fresh master and compiled with go 1.20

grubles:
Do you have `allow-deprecated-apis=false` set in your CLN config?

sat-hoarder:
I have **allow-deprecated-apis=true** set in my config

grubles:
`interface {} is float64, not string` 

Do you have `bitcoin-rpcport` set in your CLN config? If so can you paste the line for us? It appears CLN is providing PeerSwap with the wrong type (float64 instead of string).

sat-hoarder:
It was only logging INFO events.  I commented out just the bitcoin-rpcport and now peerswap loads and doesn't die.  I'm happy to post the INFO lines if you'd like

grubles:
Ok I reproduced where the plugin is killed after trying to start but not the original panic yet. I suspect the problem might be because CLN 23.08 changed the `listconfigs` API to return a `configs` subobject instead of a dict. https://github.com/ElementsProject/lightning/pull/6243

PR #226, opened 2023-08-24
cln: bitcoin-rpcport is now an int in CLN 23.08

### LINKED ISSUE #225: PeerSwap doesn't start if using CLN 23.08 and if bitcoin-rpcport is set in CLN config
When running CLN 23.08, if `bitcoin-rpcport` is set in the CLN config, PeerSwap fails to start: 

```
2023-08-24T14:49:01.937Z INFO    plugin-peerswap: PeerSwap starting up with commit 19eeea491fe39cad78302096b6b2bd922af61be8
2023-08-24T14:49:01.937Z INFO    plugin-peerswap: DB version: v0.2, Protocol version: 3
2023-08-24T14:49:01.937Z INFO    plugin-peerswap: Using data dir: /home/user/.lightning/signet
2023-08-24T14:49:01.941Z INFO    plugin-peerswap: Killing plugin: exited during normal operation
```

Commenting `bitcoin-rpcport` out fixes it and PeerSwap is able to start. 

CLN 23.08 introduced a change to `listconfigs`. Could this be the root cause?

grubles:
`bitcoin-rpcport` is now an integer in CLN 23.08. Go unmarshals JSON numbers into a `float64` by default, so we need to convert that to a `uint` to make PeerSwap happy. 

Should fix #225.

### DIFF
--- a/clightning/config.go
+++ b/clightning/config.go
@@ -7,7 +7,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"strconv"
 	"strings"
 
 	"github.com/elementsproject/glightning/glightning"
@@ -273,11 +272,7 @@ func BitcoinFallbackFromClnConfig(client *ClightningClient) Processor {
 					}
 					if v, ok := plugin.Options["bitcoin-rpcport"]; ok {
 						if v != nil {
-							port, err := strconv.Atoi(v.(string))
-							if err != nil {
-								return nil, err
-							}
-							c.Bitcoin.RpcPort = uint(port)
+							c.Bitcoin.RpcPort = uint(v.(float64))
 						}
 					}
 				}
--- a/clightning/config.go
+++ b/clightning/config.go
@@ -9,7 +9,6 @@ import (
 	"path/filepath"
 	"strings"
 	"strconv"
-	"reflect"
 	"github.com/elementsproject/glightning/glightning"
 	"github.com/elementsproject/peerswap/log"
 	"github.com/pelletier/go-toml/v2"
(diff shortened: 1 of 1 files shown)


clightning/config.go
```diff
@@ -271,11 +271,16 @@ func BitcoinFallbackFromClnConfig(client *ClightningClient) Processor {
...
-								return nil, err
+							// detect if type is string (CLN < v23.08)
+							if reflect.TypeOf(v).String() == "string" {
+								port, err := strconv.Atoi(v.(string))
+								if err != nil {
+									return nil, err
+								c.Bitcoin.RpcPort = uint(port)
+								}
+							} else {
+								c.Bitcoin.RpcPort = uint(v.(float64))
```
nepet, line 282: Nice that you also thought about backwards compatibility!
Instead of reflection (which is more powerful but also requires more memory and is slower) we could take the direct and idiomatic way, using type assertions. This way we could use a "type switch" that could also tell us about cases we missed or that we have not covered on future changes of external apis.
```suggestion
							// detect if type is string (CLN < v23.08)
							switch p := v.(type) {
							case string:
							        port, err := strconv.Atoi(p)
								if err != nil {
									return nil, err
							        }
							        c.Bitcoin.RpcPort = uint(port)
							case float64:
							        c.Bitcoin.RpcPort = uint(p)
							default:
							        return nil, fmt.Errorf("Bitcoind rpcport type %T not handled", v)
							}
```

nepet (changes requested):

nepet:
If I got it right, this should not only fix #225 but also #224?

grubles:
Let's go with your suggestion. Yes 224 should be fixed as well!

nepet (approved):
:+1:

Issue #109, opened 2022-07-15
All timestamps should be stored as UTC?


wtogami:
listpeers example:
```
   {
      "created_at": "2022-07-13 19:23:15 +0200 CEST",
   }
```

I'm guessing PeerSwap necessarily stores timestamps here and elsewhere for later processing and timeouts? 

Any timestamps stored by PeerSwap should be in the UTC timezone. (Also known as GMT or Z.)

When CLN displays any times it always prints them like this where Z is Zulu, same as UTC.
```
               "state_changes": [
                  {
                     "timestamp": "2021-05-23T07:34:16.473Z",
                     "old_state": "CHANNELD_AWAITING_LOCKIN",
                     "new_state": "CHANNELD_NORMAL",
                     "cause": "user",
                     "message": "Lockin complete"
                  }
```

For the sake of simplicity we shouldn't bother to convert timezones when displayed. Other LN related software already outputs times as UTC and sysadmins would be smart to just make the entire OS the same timezone.

https://github.com/ElementsProject/peerswap/blob/master/docs/upgrade.md
There is a tiny chance of problems during transition of existing swaps to this new timestamp. People should be told to follow the upgrade guide where they are sure no swaps are pending before upgrading.

nepet:
This is resolved by 3b649e3cf952f26c6b06653178bd14b1fedb4c11 and can be closed as soon as PR #131 is merged.

PR #229, opened 2023-08-30
delete the comment of old issue link

YusukeShimizu:
This comment seems to be about issue https://github.com/elementsproject/peerswap/issues/110, not https://github.com/elementsproject/peerswap/issues/109.  
https://github.com/elementsproject/peerswap/issues/110 is already closed, but I think the premium feature specifications have not been determined.

I think it would be better to reopen https://github.com/elementsproject/peerswap/issues/110 or create a separate issue for the premium feature.

If a separate issue is created, I'll re-modify the link to a new one.

### DIFF
--- a/swap/actions.go
+++ b/swap/actions.go
@@ -366,7 +366,6 @@ func (c *CreateSwapOutFromRequestAction) Execute(services *SwapServices, swap *S
 		return swap.HandleError(err)
 	}
 
-	// todo replace with premium estimation https://github.com/elementsproject/peerswap/issues/109
 	openingFee, err := wallet.GetFlatSwapOutFee()
 	if err != nil {
 		swap.LastErr = err


wtogami:
You're correct #109 was incorrect. But #110 I think is also incorrect. There's two separate components of the fee.

* The flat fee is prepayment for the requester to pay the onchain costs of the proposed swap.
*  The missing part is the premium which is a proportional fee for the requester to pay for opportunity cost of capital in (usually) buying incoming channel capacity.

YusukeShimizu:
There seems to be missing part for onchain cost.
GetFlatSwapOutFee only seems to add the `opening-tx` costs.
What about a refund of the on-chain funds?(`claim_by_csv_tx`)

I assume the following pattern
* onchain costs
    * `opening-tx` and `claim_by_csv_tx`
* premium

nepet:
@YusukeShimizu This seems to be correct.
As of now, we only take the `opening-tx` fee into account. We always tried to keep the trust overhead (fee-invoice) low and meet somewhere in the middle. We assume allowlisted peers to pay the invoice and claim the on-chain htlc, as they already committed themselves by paying the fee-invoice. 
The `opening-tx` on the other hand is always paid for by the party that broadcasts the tx so the swap fee as it is now covers the definite on-chain fee. 
Eventually we might switch to a deposit scheme where the `claim-by-csv-tx` is also covered to some extend, but where this part of the "fee" is returned via the on-chain tx if the initiator pays the invoice and claims by preimage.
Premium is on top of that and not security related.

I think you are right changing #109 to #110 in the comment, but the comment is somewhat misleading anyways. It should say something like: _Add premium once we have an idea on how to do that (mainly, how to negotiate it)_

nepet:
By the way, this is an ancient comment that dates back to a time when we did not distinguish between premium and "upfront-fee" in terminology. You may consider removing the comment altogether.

PR #230, opened 2023-08-30
config: add function to detect allow-deprecated-apis=false in CLN config

### LINKED ISSUE #228: we should warn if CLN users set allow-deprecated-apis=false
CLN will deprecate parts of its API and these changes will have the potential to negatively affect the function of PeerSwap. The CLN team gives developers and users some time before fully removing the deprecated parts, and include a config option `allow-deprecated-apis=bool` to toggle the deprecated API parts on and off for testing purposes. 

We should detect the config option and print a warning.

grubles:
This adds a function `CheckForDeprecatedApiConfig()` to `clightning/config.go` to check if `allow-deprecated-apis` is set to `false` in the CLN config. This is because setting that option false turns off deprecated, but not yet fully removed, API fields that PeerSwap may rely on. If we detect it is set `false` then we print a warning and exit the plugin to be safe.

Resolves #228

### DIFF
--- a/clightning/config.go
+++ b/clightning/config.go
@@ -117,7 +117,7 @@ func CheckForDeprecatedApiConfig(client *ClightningClient) Processor {
 		x := listConfigResponse["configs"].(map[string]interface{})["allow-deprecated-apis"]
 		z := x.(map[string]interface{})["value_bool"]
 		if z == false {
-			log.Infof("WARNING: allow-deprecated-apis=false detected in CLN config. Exiting")
+			log.Infof("WARNING: allow-deprecated-apis=false detected in CLN config. Exiting. More info: https://github.com/ElementsProject/peerswap/issues/232")
 			time.Sleep(1 * time.Second)
 			os.Exit(1)
 		}
--- a/clightning/config.go
+++ b/clightning/config.go
@@ -7,12 +7,13 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"strings"
 	"strconv"
+	"strings"
+	"time"
+
 	"github.com/elementsproject/glightning/glightning"
 	"github.com/elementsproject/peerswap/log"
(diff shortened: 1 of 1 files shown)


clightning/config.go
```diff
@@ -95,6 +96,35 @@ func SetPeerswapPaths(plugin *glightning.Plugin) Processor {
...
 
+// CheckForDeprecatedApiConfig tries to detect if allow-deprecated-apis is false
+// in the CLN config. If it is set false, we print a warning and exit because
+// deprecated CLN API fields might break PeerSwap.
+func CheckForDeprecatedApiConfig(client *ClightningClient) Processor {
+	return func(c *Config) (*Config, error) {
+		conf, err := client.glightning.ListConfigs()
+			if err != nil {
+				return nil, err
+			}
```
YusukeShimizu, line 107: Nit: I think one extra indent is added.
How about the following?

```suggestion
		if err != nil {
			return nil, err
		}
```

clightning/config.go
```diff
@@ -95,6 +96,35 @@ func SetPeerswapPaths(plugin *glightning.Plugin) Processor {
...
+		data, err := json.Marshal(conf)
+			if err != nil {
```
YusukeShimizu, line 109: ditto: https://github.com/ElementsProject/peerswap/pull/230#discussion_r1313717712

clightning/config.go
```diff
@@ -95,6 +96,35 @@ func SetPeerswapPaths(plugin *glightning.Plugin) Processor {
...
+				return nil, err
+			}
+		var listConfigResponse map[string]interface{}
+		err = json.Unmarshal(data, &listConfigResponse)
+			if err != nil {
```
YusukeShimizu, line 114: ditto: https://github.com/ElementsProject/peerswap/pull/230#discussion_r1313717712

PR #233, opened 2023-09-01
remove poll disconnected peer error message 

### LINKED ISSUE #223: silence useless poll error
```
2023-08-23T20:49:32.931Z DEBUG   plugin-peerswap: poll_service: could not send msg to 0349652d0774034aa2945ecb5f81acc55c3e362b7d87be20752daf53c13f4254b3: -32600:No such peer: 0349652d0774034aa2945ecb5f81acc55c3e362b7d87be20752daf53c13f4254b3
```

We don't need to log this failure when we poll a peer that is not connected.

This is most visible during CLN startup when you have many peers.

grubles:
This can get spammy for larger nodes. Fixes #223.


wtogami:
Oh. I didn't know we already prevent it from logging more than once. Sorry let's keep it the way it already is. My bad.

Issue #81, opened 2022-06-27
Sanity check channel balance prior to swap


wtogami:
Prior to starting a swap, and again prior to sending, check the channel's max htlc limit to verify that the requested swap would actually be possible. Print informative error messages otherwise.

Related: #73

nepet:
## Why
So this issue is a little bit paranoid but we want to do everything we can to avoid canceled swaps, especially swap-out swaps as the `responder` is in charge of the on-chain part.
The following assumes a swap-out but same applies symmetrically for a swap-in:
Currently we only check that we have enough local balance (`our_amount_msat`) to perform the swap. This check is not strong enough as we could have set a `htlc_maximum_msat` that is below our local balance.

## Solution
We want to check that `min(htlc_maximum_msat, our_amount_msat) > swap_amt_msat + fees`
Best place to check that would be after we received the `swap_out_agreement` that contains the "fee_invoice". We may still want to perform the same check without the fees before we initiate a swap.

wtogami:
Nothing is currently broken without this but this would be a nice to have because it prevents events that would immediately fail from adding things to the database.

PR #235, opened 2023-09-05
Add additional sanity checks before starting a swap

nepet:
We add an additional check that we can spend/receive the swap amount over the desired channel before initiating a swap.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,8 @@ test-bitcoin-cln: test-bins
 	'Test_ClnCln_Bitcoin_SwapOut|'\
 	'Test_ClnCln_Bitcoin_SwapIn|'\
 	'Test_ClnLnd_Bitcoin_SwapOut|'\
-	'Test_ClnLnd_Bitcoin_SwapIn)'\
+	'Test_ClnLnd_Bitcoin_SwapIn|'\
+	'Test_ClnCln_ExcessiveAmount)'\
 	 ./test
 .PHONY: test-bitoin-cln
 
--- a/clightning/clightning.go
+++ b/clightning/clightning.go
@@ -208,23 +208,38 @@ func (r ListPeerChannelsRequest) Name() string {
 	return "listpeerchannels"
 }
 
+// ListPeerChannelsResponse is the response to the `listpeerchannels` rpc
+// method. This struct is incomplete, see
+// `https://docs.corelightning.org/reference/lightning-listpeerchannels` for a
+// full list of fields that are returned.
+type ListPeerChannelsResponse struct {
+	Channels []PeerChannel `json:"channels"`
+}
+
+type PeerChannel struct {
+	PeerConnected  bool              `json:"peer_connected"`
+	State          string            `json:"state"`
+	ShortChannelId string            `json:"short_channel_id,omitempty"`
+	TotalMsat      glightning.Amount `json:"total_msat,omitempty"`
+	ToUsMsat       glightning.Amount `json:"to_us_msat,omitempty"`
+	ReceivableMsat glightning.Amount `json:"receivable_msat,omitempty"`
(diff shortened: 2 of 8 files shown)


YusukeShimizu:
Since there are similarities with `CheckChannel` function, I think it would be good if `CheckChannel` could be eliminated and integrated into checks using `SpendableMsat`.

The following may be the scope of https://github.com/ElementsProject/peerswap/issues/81, but I mention it because I am thinking of working on it.

I think it would be nice to be able to check `htlc_maximum_msat` as well.
it would be nice to check `SpendableMsat` after we received the swap_out_agreement that contains the "fee_invoice". 

I would appreciate your feedback.

wtogami:
Is it common for LND nodes to adjust `htlc_maximum_msat` dynamically?

nepet:
> Since there are similarities with `CheckChannel` function, I think it would be good if `CheckChannel` could be eliminated and integrated into checks using `SpendableMsat`.

True, there is indeed some room for refactoring and improvement. I will refactor said methods and the interface once I know that the pr fixes @grubles problem with the swap amounts.

> The following may be the scope of #81, but I mention it because I am working on it.
> 
> I think it would be nice to be able to check `htlc_maximum_msat` as well.

We can only ever give an estimate on what is spendable. For cln we can use `spendable_msat` if it is set. Otherwise we have to deal with a fallback. It is definitely a good idea to check `htlc_maximum_msat` as long as we don't split the payment.

> it would be nice to check `SpendableMsat` after we received the swap_out_agreement that contains the "fee_invoice".

I would rather check before we start the swap at all, once we are at the point where we want to pay an invoice we can just do so and act accordingly if it fails.

nepet:
Oh sorry, maybe I misunderstood, of course it would be nice to check if we can pay the `whole` swap amount before paying the `fee_invoice` ^^

YusukeShimizu:
> I will refactor said methods and the interface once I know that the pr fixes @grubles problem with the swap amounts.

I understood.
LGTM

> it would be nice to check if we can pay the whole swap amount before paying the fee_invoice 

Yes, this is what I wanted to tell you.

I will implement it as a response to https://github.com/ElementsProject/peerswap/issues/81.

wtogami:
This PR seems to do the right thing.

> I will refactor said methods and the interface once I know that the pr fixes @grubles problem with the swap amounts.

Here it sounds that Peter will refactor to check `htlc_maximum_msat`.

> I will implement it as a response to https://github.com/ElementsProject/peerswap/issues/81.

But Yusuke said he can implement it after this is merged.

Let's merge this now since it works then you two discuss who takes the next step?

nepet:
Let me clean this up a bit so that we can merge it as is (without `htlc_maximum_msat`). I am happy to pass this to @YusukeShimizu for further steps.

Is this okay for you @YusukeShimizu, @wtogami?

YusukeShimizu:
Yes, I am OK.

I will implement the following two features.
1. check if we can pay the whole swap amount before paying the fee_invoice
2. make htlc_maximum_msat be taken into account when calculating SpendableMsat

wtogami:
https://github.com/ElementsProject/peerswap/actions/runs/6177482756/job/16768697864?pr=235
Test failure

grubles:
Unpatched CLN node impossible swap-out is rejected by remote patched LND peer. Impossible swap-in is rejected locally.

```
{                                                                                                                                                                                                                  
   "code": -1,                                                                                                                                                                                                     
   "message": "swap canceled, reason: from the LND peer: exceeding receivable amount_msat: 236942513000"                                                                                                           
} 
```

grubles:
Don't merge yet. I want to test a little bit more with different configurations (a CLN <> CLN channel and a LND <> LND channel).

grubles:
Successfully tested with two LND nodes (one unpatched, one patched).

PR #241, opened 2023-09-11
Wait for swap in State_SwapInSender_AwaitAgreement if it exists

YusukeShimizu:
Fixes flake https://github.com/ElementsProject/peerswap/actions/runs/6104788193.
If there is a swap in the State_SwapInSender_AwaitAgreement state with cancel message, it is likely to complete after a certain period of time, so wait.


nepet:
Ack a6636a0

PR #242, opened 2023-09-12
test: the amount of the fee invoice is taken into account when testing

YusukeShimizu:
This is a fix for a testing bug and the resulting reduction in testing time.
I expect to shorten the test time by approximately 7 minutes.

Claim coop testing of swap out time is dominating.  
The reason seems to be that the channel taker balance check is stuck in an infinite loop until time out.

In the case of swap out, the amount of the fee invoice needs to be taken into account.

### DIFF
--- a/test/testcases.go
+++ b/test/testcases.go
@@ -44,6 +44,9 @@ func coopClaimTest(t *testing.T, params *testParams) {
 	//
 	// Move local balance from taker to maker so that the taker does not
 	// have enough balance to pay the invoice and cancels the swap coop.
+	feeInvoiceAmt, err := params.makerNode.GetFeeInvoiceAmtSat()
+	require.NoError(err)
+
 	moveAmt := (params.origTakerBalance - params.swapAmt) + 100
 	inv, err := params.makerNode.AddInvoice(moveAmt, "shift balance", "")
 	require.NoError(err)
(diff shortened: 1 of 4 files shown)


YusukeShimizu:
ack

nepet:
Very nice catch!
How about this change to remove the upper-lower bound check since we now know the correct amount? Checking the bounds can still lead to races with Lnd as we might catch an intermediary channel state that has an actually lower local balance than it should have.

```diff
diff --git a/test/testcases.go b/test/testcases.go
index 21a093c..22ad827 100644
--- a/test/testcases.go
+++ b/test/testcases.go
@@ -47,7 +47,7 @@ func coopClaimTest(t *testing.T, params *testParams) {
        feeInvoiceAmt, err := params.makerNode.GetFeeInvoiceAmtSat()
        require.NoError(err)
 
-       moveAmt := (params.origTakerBalance - params.swapAmt) + 100
+       moveAmt := (params.origTakerBalance - feeInvoiceAmt - params.swapAmt) + 1
        inv, err := params.makerNode.AddInvoice(moveAmt, "shift balance", "")
        require.NoError(err)
 
@@ -59,8 +59,7 @@ func coopClaimTest(t *testing.T, params *testParams) {
        err = testframework.WaitFor(func() bool {
                setTakerFunds, err = params.takerNode.GetChannelBalanceSat(params.scid)
                require.NoError(err)
-               return params.origTakerBalance-moveAmt-10-feeInvoiceAmt < setTakerFunds &&
-                       setTakerFunds < params.origTakerBalance-moveAmt+10-feeInvoiceAmt
+               return setTakerFunds == params.swapAmt-1
        }, testframework.TIMEOUT)
        require.NoError(err)
```

YusukeShimizu:
Thanks for the suggestion.
I agree and have made the change. https://github.com/ElementsProject/peerswap/pull/242/commits/9c2afaf89562efe1bb251444e896fb8d00e94d9b

nepet:
ACK 9c2afaf89562efe1bb251444e896fb8d00e94d9b

PR #245, opened 2023-09-15
Docs: Fix sequence swapin

YusukeShimizu:
Removed the bloadcast of claim tx on the responder side from the coop close, since the bloadcast of claim tx is done by initiator.

The receiving node of the `coop_close message` broadcasts the claim_transaction as described in [the specification](https://github.com/ElementsProject/peerswap/blob/master/docs/peer-protocol.md#the-coop_close-message).


docs/mmd/swap-in-sequence.md
```diff
@@ -20,7 +20,6 @@ B-->>A: pay claim invoice
 Note over B: broadcast claim tx
 else refund cooperatively
 Note over B: sign coop claim tx
```
nepet, line 22: This line seems also useless, the `coop_close` message contains the key so that the maker can construct the `claim_tx` himself.

nepet (changes requested):
Nice catch again! It is really good to have a "new" pair of eyes that scans through things in detail, thanks!

docs/mmd/swap-in-sequence.md
YusukeShimizu: Yes, I agree.
Deleted. https://github.com/ElementsProject/peerswap/pull/245/commits/d2e6a312d6e786cb09496f53724f9a89cd26e0d0

nepet (approved):
Ack fcc8b94c9548c35122186aa348e47ced62328c41

PR #246, opened 2023-09-21
Sanity check max htlc amount msat

YusukeShimizu:
Fixes https://github.com/ElementsProject/peerswap/issues/81

* check the channel's max htlc limit
Added logic to check the maximum htlc limit set per node for each channel.
In cln, max htlc amt is not affect, so checks are only performed when the sender is lnd.
receivable amount depends on whether the node is a cln or a lnd, so max htlc amt is not taken into account.
Also, since the max htlc limit is not always set reliably, the check is skipped if it is not set.

* add sanity check of spendable amt at `swap_out_agreement `

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,8 @@ test-bitcoin-lnd: test-bins
 	'Test_LndLnd_Bitcoin_SwapOut|'\
 	'Test_LndLnd_Bitcoin_SwapIn|'\
 	'Test_LndCln_Bitcoin_SwapOut|'\
-	'Test_LndCln_Bitcoin_SwapIn)'\
+	'Test_LndCln_Bitcoin_SwapIn|'\
+	'Test_LndLnd_ExcessiveAmount)'\
 	 ./test
 	${INTEGRATION_TEST_ENV} go test $(INTEGRATION_TEST_OPTS) ./lnd
 .PHONY: test-bitcoin-lnd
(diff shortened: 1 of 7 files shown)


grubles:
Tested by manually setting the max HTLC to the absolute minimum for both CLN and LND. Currently CLN does not apply the max HTLC to payments using `sendpay`, so setting it doesn't matter here.

With the max HTLC set to 1000msat on the LND side and with LND as a swap-in receiver, the swap is canceled:

```
$ lightning-cli --signet peerswap-swap-in 140490x1x0 200000000 lbtc
{
   "code": -1,
   "message": "swap canceled, reason: from the LND peer: exceeding spendable amount_msat: 1000"
}
```

With LND as a swap-out sender, the swap is also canceled:

```
$ pscli swapout --sat_amt 240000000 --channel_id 154470388586315776 --asset lbtc
2023/09/25 08:57:42 rpc error: code = Unknown desc = exceeding spendable amount_msat: 1000
```

Restoring the max HTLC amount for LND allows the swap to occur.

wtogami:
> In cln, max htlc amt is not affect, so checks are only performed when the sender is lnd.

Can max htlc be queried prior to beginning a CLN swap? If so this could go in another PR later. I consider this to be a low priority.

Merging this for now.

YusukeShimizu:
Yes, it is possible.
I will respond with another PR.

PR #250, opened 2023-10-09
swap: bump protocol version to 4

grubles:
Bumps the protocol version from 3 to 4. 

MERGE THIS AS THE LAST THING BEFORE RELEASE.


wtogami:
* Does the spec need to be updated to match the code protocol version?
* We merge this version bump when the release notes are ready.

wtogami:
Peter said the protocol specs need to be updated to match this version bump.

nepet:
I just bumped the version of the spec that is in the docs here. We still have https://github.com/ElementsProject/peerswap-spec which we may want to include as a `git submodule` to replace the spec in this repository.

PR #252, opened 2023-10-22
compile protobufs in docker

YusukeShimizu:
This is an improvement in the development experience.
No behavior changes.

This commit aims to make it easier to compile protobuf.
No need to have whole set of binaries and libraries all pinned to specific versions.


nepet (approved):
Nice work! even though I would prefer nix for reproducibility issues ;)
This will be a major improvement for new devs.
ACK 98b613b5ffe712e5548852ea5b1fddc01f297292

PR #253, opened 2023-11-08
swap: automatically add to the suspicious list If csv recovery

YusukeShimizu:
If csv recovery occurs, add to the suspicious list.

Csv recovery occurs only in rare cases.
When such a loss occurs, automatically add the peer to the bad list.
That way, such rare losses will not be repeated.


nepet (approved):
ACK 023a2a5d477f917d73fb0bab1c362f0bc65479c8

PR #255, opened 2023-11-14
swap: cancel swap on recovery

YusukeShimizu:
Fixes [discord support bug report](https://discord.com/channels/905126649224388629/905127434779766785/1154271226756349952).

Cancel swap on recovery during State_SwapOutSender_AwaitAgreement.
Previously, when recovery on `State_SwapOutSender_AwaitAgreement` occurred, it would hang.

The receiver side is canceled when recovery occurs at `State_SwapOutReceiver_CreateSwap`,
so this change does not affect the receiver side.


nepet:
Good catch!
Ack baee1f42d959e70ae6b47cd6029a36dc7757dfcf

PR #258, opened 2023-11-20
liquid: return more accurate errors in createSpendingTransaction

jamesdorfman:
Currently, this function returns `pset.ErrGenerateSurjectionProof` in many places where there is not actually a surjection proof error. This made debugging very difficult. 

This PR updates the errors to be more accurate. I am open to adding more information to these errors, aside from just returning `err`. Let me know your thoughts.


nepet:
I am okay with returning the `err` only for now. When this becomes a problem in the future we can wrap the error to include more information.
Ack addfc7b3ca5f6388be8d2cd5a2edc1e69adfa95c

PR #269, opened 2024-01-10
cln: remove wumbo check because CLN >=v23.11 defaults to enabling it

### LINKED ISSUE #261: Latest CLN defaults to wumbo now. PeerSwap's wumbo / large-channels check is broken.
```
$ lightning-cli --signet peerswap-swap-out 140490x1x0 246942696 lbtc                                                                                      
{
   "code": -1,
   "message": "swap amount is 246942696000: need to enable option '--large-channels' to swap amounts larger than 2^32 msat"
}
```

```
$ cat ~/.lightning/config
signet
plugin=/home/user/.lightning/signet/plugins/peerswap
daemon
log-file=logs
log-level=debug
```

`large-channels` is missing, so PeerSwap considers it turned off and won't allow the swap to continue past that check. 

Even adding `large-channels` to the CLN config does not help.

Commenting this line out allows the swap to continue: https://github.com/ElementsProject/peerswap/blob/c22211a9ea6aad15af32266901d620afd1ad2c5f/clightning/clightning.go#L187

grubles:
Latest CLN ignores `large-channels` / `wumbo` config options and does not pass them to PeerSwap. PeerSwap checks for this config option and cancels large swaps if it is missing. This PR removes `ClnMaxPaymentSizeMsat` and the check since we don't need it anymore for CLN >=23.11.

Fixes #261

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,6 @@ test-misc-integration: test-bins
 	'Test_RestoreFromPassedCSV|'\
 	'Test_Recover_PassedSwap_BTC|'\
 	'Test_Recover_PassedSwap_LBTC|'\
-	'Test_Wumbo|'\
 	'Test_ClnConfig|'\
 	'Test_ClnPluginConfigFile|'\
 	'Test_ClnPluginConfigFile_DoesNotExist|'\
--- a/clightning/clightning.go
+++ b/clightning/clightning.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/hex"
-	"encoding/json"
 	"errors"
 	"fmt"
 	log2 "log"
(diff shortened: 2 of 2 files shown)


YusukeShimizu (approved):
ack

Makefile
```diff
@@ -137,7 +137,6 @@ test-misc-integration: test-bins
 	'Test_RestoreFromPassedCSV|'\
 	'Test_Recover_PassedSwap_BTC|'\
 	'Test_Recover_PassedSwap_LBTC|'\
-	'Test_Wumbo|'\
```
nepet, line 140: Is the wumbo test also used in a different place or can we remove the test completely from the `tests`?

nepet (commented):
ACK 23f3ac5bbb5858f65ea20d46760b3ee5a4f3502e

Makefile
grubles: I didn't feel 100% comfortable removing `wumbo_test.go` too just in case it was used elsewhere. Pretty certain it is fine to remove it though.

PR #263, opened 2023-12-19
bump nixpkgs and add --developer flag to cln test config

grubles:
Bumps daemons to the following:

cln v23.11.1
lnd v0.17.0-beta
bitcoin v25.1
elements v23.2.1

Also adds `--developer` flag to the testframework CLN config since `--enable-developer` is removed in the latest CLN. 

Fixes #249

PR #271, opened 2024-01-15
test: force fee update

YusukeShimizu:
Related to https://github.com/ElementsProject/peerswap/pull/263

# What
Implemented ForceFeeUpdate.
New feerate allows us to force a fee update by restarting the node.

# Why
This is in response to GetRoute returning `no route` on a stuck channel since v23.11 of cln.
The fee rate can be updated to make a channel stuck after it has been opened.


nepet (approved):
ACK d0bb5243541129e3c65f95fdb71c98b0786492a7

Issue #96, opened 2022-07-11
[feature request] Dedicated onchain wallet for PeerSwap


openoms:
Moving the discussion from #88.

Idea:
Have a separate pool of funds used for PeerSwap, different from the CLN / LND onchain wallet on the same instance.
The LN wallet is already used for:
* funding channels 
* receiving funds from closed channels (force closes always return to the internal wallet).
* CLN:  liquidity ads / dual funded channels

The ability to keep the funds used for different purposes would be good for privacy and could also simplify the accounting.

Possibly related: https://github.com/GaloyMoney/charts/issues/1208

Issue #88, opened 2022-07-06
CLN configuration help using bitcoin RPC password and username 


openoms:
Running CLN with the plugin from the latest PeerSwap commit using RPC password and username.
The pluging exits right after starting. What the problem could be?

logs:
```
INFO    plugin-bcli: bitcoin-cli initialized and connected to bitcoind.
INFO    lightningd: --------------------------------------------------
INFO    lightningd: Server started with public key 03b2b374a259d2ec8ca46f7db079fbb12b6e307dbb319b1d2f911500e332e41d41, alias BIZARREARK (color #03b2b3) and lightningd v0.11.2
INFO    plugin-peerswap-plugin: PeerSwap Initialized, running PeerSwap commit baf6e4c38d16dcd922f94e777bcd892db5b0bc5f
INFO    plugin-peerswap-plugin: Liquid swaps disabled
DEBUG   plugin-peerswap-plugin: connecting with RPC_USER, RPC_PASSWORD to :0
INFO    plugin-peerswap-plugin: Killing plugin: exited during normal operation
```

config:
```
# lightningd configuration for bitcoin testnet
network=testnet
log-file=cl.log
log-level=debug:plugin-peerswap-plugin
plugin-dir=/home/bitcoin/tcl-plugins-enabled

# Tor settings
proxy=127.0.0.1:9050
bind-addr=127.0.0.1:19736
addr=statictor:127.0.0.1:9051/torport=19736
always-use-proxy=true

peerswap-bitcoin-rpcuser=RPC_USER
peerswap-bitcoin-rpcpassword=RPC_PASSWORD
peerswap-elementsd-enabled=false
peerswap-bitcoin-rpchost=localhost
peerswap-bitcoin-rpcport=18332
```

openoms:
Tried with `peerswap-bitcoin-rpchost=127.0.0.1` and with the line removed, but the output remains the same:
```
DEBUG   plugin-peerswap-plugin: connecting with RPC_USER, RPC_PASSWORD to :0
INFO    plugin-peerswap-plugin: Killing plugin: exited during normal operation
```

openoms:
IPv6 was not used on any related interface, but in any case disabled everywhere and also restarted, but no change,

grubles:
Not sure how valuable this feedback is but I've run into this before too. I had a few different parameters set so I couldn't determine which was causing the plugin to crash/exit. On a separate box with nothing explicitly set -- basically using whatever peerswap expects -- worked fine.

[...]

PR #283, opened 2024-02-13
Set label

YusukeShimizu:
Ensure that all cln/liquid transactions related to peerswaps are identifiable by the label.
This makes it easier to audit the transactions.
if it fails, it is logged and subsequent processing is continued.

https://github.com/ElementsProject/glightning/compare/master...YusukeShimizu:glightning:SetLabel?expand=1


go.mod
```diff
@@ -151,3 +151,5 @@ require (
 
 // This fork contains some options we need to reconnect to lnd.
 replace github.com/grpc-ecosystem/go-grpc-middleware => github.com/nepet/go-grpc-middleware v1.3.1-0.20220824133300-340e95267339
+
+replace github.com/elementsproject/glightning => github.com/yusukeshimizu/glightning v0.0.0-20240213204501-f17161a057dc
```
YusukeShimizu, line 155: [todo]remove this once glightning is updated

wtogami:
Let's do this ...

1. For now let's add labels for peerswap LND only. LND already provides the needed RPC commands. Document the limitation that only peerswap LND has onchain labels for now.
1. We will not add https://github.com/ElementsProject/glightning/pull/12 as this would enable CLN peerswap to label only elementsd. As noted there glightning is already overloaded with too many things which I consider to be a layering violation.
1. Next CLN peerswap should talk to an optional secondary bitcoind wallet directly like [#96](https://github.com/ElementsProject/peerswap/issues/96) or talk directly to elementsd or LWK to set labels. Do this in another PR later.

YusukeShimizu:
> For now let's add labels for peerswap LND only. LND already provides the needed RPC commands. Document the limitation that only peerswap LND has onchain labels for now.

YusukeShimizu:
> 今のところ、peerswap LND のためだけにラベルを追加しましょう。LNDはすでに必要なRPCコマンドを提供している。今のところ、peerswap LNDだけがonchainラベルを持つという制限を文書化してください。

OK、それを文書化するPRを作成します。

> これはCLN peerswapがelementsdのみにラベルを付けることを可能にするので、https://github.com/ElementsProject/glightning/pull/12。

このPRの変更は、clnが使用されていてもbitcoindのセットラベルを呼び出します。
このbitcoindはclnが依存しているものと同じです。

> glightningはすでに多くのことを詰め込みすぎていて、レイヤー違反だと思います。

そうですね。
各デーモン(bitcoin,elementsd,cln)などを分けるべきかもしれません。

> 次のCLNのpeerswapは、https://github.com/ElementsProject/peerswap/issues/96 のようにオプションのセカンダリbitcoindウォレットと直接話すか、ラベルを設定するためにelementsdまたはLWKと直接話すべきです。これは後で別のPRで行う。

cln/lndが使用するデーモンと別のデーモンに接続するオプションを提供することは可能でしょう。
これは後で対応するべきだと理解した。

YusukeShimizu:
> For now let's add labels for peerswap LND only. LND already provides the needed RPC commands. Document the limitation that only peerswap LND has onchain labels for now.

OK, I will create a PR to document it.

> We will not add https://github.com/ElementsProject/glightning/pull/12 as this would enable CLN peerswap to label only elementsd.

This PR change calls the set label of bitcoind even if cln is used.
This bitcoind is the same as the one cln depends on.

> As noted there glightning is already overloaded with too many things which I consider to be a layering violation.

I agree.
Maybe we should separate each daemon (bitcoin,elementsd,cln) etc.

> Next CLN peerswap should talk to an optional secondary bitcoind wallet directly like https://github.com/ElementsProject/peerswap/issues/96 or talk directly to elementsd or LWK to set labels. Do this in another PR later.

It would be possible to offer the option to connect to the daemon used by cln/lnd and another daemon.
I understood that this should be addressed later.

wtogami:
> > We will not add [ElementsProject/glightning#12](https://github.com/ElementsProject/glightning/pull/12) as this would enable CLN peerswap to label only elementsd.
> 
> This PR change calls the set label of bitcoind even if cln is used. This bitcoind is the same as the one cln depends on.

CLN uses bitcoind only to read blocks/transactions. CLN has its own onchain wallet which unfortunately lacks labels entirely.

Requests like #96 have asked peerswap to offer the ability to use a different BTC onchain wallet for those who want to keep it separate from their CLN funds. Even if we add this as an option it would be confusing to have setlabel when it can't work with the default CLN onchain wallet.

That said maybe we should include [ElementsProject/glightning#12](https://github.com/ElementsProject/glightning/pull/12) limited to elementsd for now since it would work for our default L-BTC wallet. Thoughts @nepet ?

Related question: Does our CLN peerswap currently communicate entirely via CLN RPC or does it also connect to bitcoind and elementsd directly?

wtogami:
> That said maybe we should include https://github.com/ElementsProject/glightning/pull/12 limited to elementsd for now since it would work for our default L-BTC wallet. Thoughts @nepet ?

For now let's go ahead with this part. Please note the CLN bitcoind setlabel is currently impossible due to the lack of labels.

wtogami:
> Related question: Does our CLN peerswap currently communicate entirely via CLN RPC or does it also connect to bitcoind and elementsd directly?

How does this currently work?

YusukeShimizu:
My understanding is as follows

## cln
In the case of cln, peerswap starts as a cln plugin, but peerswap also has its own jsonrpc connection to bitcoind for onchain communication.
Thus, we can call `SetLabel` directly.

The default is to connect to the same bitcoind as the bitcoind that cln connects to, but it is also possible to set up a different bitcoind.

If liquid is enabled, peerswapd calls the function via elementd jsonrpc.

```mermaid
graph LR
    subgraph clnplugin
    peerswap --> cln
end
cln --> bitcoind,etc
peerswap --> alternativebitcoind
peerswap ---> elementd
```

My understanding is that https://github.com/ElementsProject/peerswap/issues/96 is resolved with https://github.com/ElementsProject/peerswap/pull/153 for cln.([Cause](https://github.com/ElementsProject/peerswap/issues/88#issuecomment-1177801885))

## lnd
In lnd, onchain exchange is done using [Wallet](https://docs.lightning.engineering/lightning-network-tools/lnd/wallet) in lnd's grpc, and rpc with bitcoind connection to bitcoind is not provided.


```mermaid
graph LR
peerswapd --> lnd
peerswapd --> elementd
subgraph lndnode
    lnd --> bitcoind,btcd,neutrino
end
```

Therefore, for lnd, there is currently no way to set up another onchain wallet, so this is something to consider.

```mermaid
graph LR
classDef color font-style:italic,stroke:yellow

peerswapd --> lnd
peerswapd --> elementd
subgraph lndnode
    lnd --> bitcoind,btcd,neutrino
end
peerswapd --> alternativebitcoindamon:::color
```

YusukeShimizu:
When swapout is performed in cln with this PR, the label can be confirmed in bitcoind connected to peerswap as follows.

```sh
./bin/clncli peerswap-listswaps
{
   "swaps": [
      {
         "id": "085b1368810005a30610adf6ba00ce01258d2d2605798ec20663feb16023a5b9",
         "created_at": 1708562187,
         "asset": "btc",
         "type": "swap-out",
         "role": "sender",
         "state": "State_ClaimedPreimage",
         "initiator_node_id": "022d6b27d4577dad8ee17466be3e9e4af6e2ac80a600453c28f8be2ae66af8b548",
         "peer_node_id": "02496a469ebf47bce20b7c0ac661f306b84cde3bbf6860d868e117d8ff3f18062c",
         "amount": 1000000,
         "channel_id": "231x1x0",
         "opening_tx_id": "4ae28ab0e547fcf23579059a1689a9f5c91dc97c2c9e804ba0d476171085be7f",
         "claim_tx_id": "d3ceb004d9464e266a181ea4639e4e836820d5ff8fdfe19b7d846c53d46362da",
         "lnd_chan_id": 253987186081792
      }
   ]
}
```

```sh
./bin/bitcoin-cli listlabels
[
  "",
  "peerswap -- ClaimByInvoice(swap id=085b13)"
]
```

```sh
./bin/bitcoin-cli getaddressesbylabel "peerswap -- ClaimByInvoice(swap id=085b13)"
{
  "bcrt1q7y2zlrysa3qjsqz0x4m6jh6w8tun996y3cfzw3": {
    "purpose": "send"
  }
}
```

wtogami:
> In the case of cln, peerswap starts as a cln plugin, but peerswap also has its own jsonrpc connection to bitcoind for
> onchain communication.
> Thus, we can call SetLabel directly.
> In this case, the label is attached to the bitcoind that peerswap connects to.

The default CLN onchain wallet does not use bitcoind's wallet. CLN works with bitcoind disablewallet=1.

> My understanding is that #96 is resolved with #153 for cln.([Cause](https://github.com/ElementsProject/peerswap/issues/88#issuecomment-1177801885))

I'm confused how this is possible. #153 was not intended to do anything like #96. If it did so it was an accident. Supporting #96 would require lots of extra work because it would be confusing and difficult to support querying two different bitcoinds.

If CLN peerswap currently can do a swap with a secondary bitcoind wallet that is NOT INTENDED. CLN would not see those UTXO's because it is not within its own onchain wallet. The user would need to manually spend it to make it available to CLN again.

> Therefore, for lnd, there is currently no way to set up another onchain wallet, so this is something to consider.

Nobody requested this feature for LND peerswap.

YusukeShimizu:
> I'm confused how this is possible. https://github.com/ElementsProject/peerswap/pull/153 was not intended to do anything like https://github.com/ElementsProject/peerswap/issues/96. If it did so it was an accident. Supporting https://github.com/ElementsProject/peerswap/issues/96 would require lots of extra work because it would be confusing and difficult to support querying two different bitcoinds.

I was confusing the bitcoin daemon with the bitcoin wallet.
https://github.com/ElementsProject/peerswap/pull/283#issuecomment-1958464183 is different from cln's wallet, which is labeled bitcoind's wallet.

> If CLN peerswap currently can do a swap with a secondary bitcoind wallet that is NOT INTENDED. CLN would not see those UTXO's because it is not within its own onchain wallet. The user would need to manually spend it to make it available to CLN again.

Currently it is not possible to swap with multiple wallets.

Since it is not possible to label the cln wallet, I will try to set the label only on the cln elementd only, as discussed separately.

mrfelton (approved):

PR #286, opened 2024-02-29
docs: add warning about alpine linux / musl and elementsd [skip ci]

grubles:
Elementsd on musl-based Linux, such as Alpine Linux or Void Linux, requires extra compile-time flags otherwise it will crash. This adds a warning to `setup_elementsd.md`.


wtogami:
please squash to a single commit

PR #291, opened 2024-03-29
bump nixpkgs and go version

grubles:
This bumps `go` to 1.22 since 1.19 is EOL now. 

This also updates the test daemons to:

cln v24.02.1
lnd v0.17.4-beta
bitcoin v26.0

### DIFF
--- a/packages.nix
+++ b/packages.nix
@@ -1,11 +1,11 @@
 let
-# Pinning to revision c8a66e2bb84c2b9cf7014a61c44ffb13ef4317ed
-# - cln v23.11
-# - lnd v0.17.0-beta
-# - bitcoin v25.1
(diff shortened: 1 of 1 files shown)


grubles:
```
setup.go:113: lightingds[0].OpenChannel() FundChannel() 313:We would not have enough left for min-emergency-msat 25000sat
```

Anchor channels are now the default in >24.02. Looks like we need to account for the extra 250 sat requirement now when opening a channel? @YusukeShimizu

YusukeShimizu:
Yes, we need to account for the `min-emergency-msat` now when opening a channel.
I created another PR to ensure all tests are passed.
https://github.com/ElementsProject/peerswap/pull/292

Issue #290, opened 2024-03-28
Failed peerswap from 63 days ago blocking two peers doing any further swapping


TrezorHannes:
Hello everyone,

Problem statement: we have had a failed ps-swapout attempt from my side which failed, and has a state-cancelled on my end. No neither of us can do another swap with each other.

_My side_ has `"state": "State_SwapCanceled",`
```
pscli getswap --id f02f50b2317a9a8dfa8a6a8a7ccab7ec25559f2eb3401d10626d4fdc3c1442f5
{
  "swap": {
    "id": "f02f50b2317a9a8dfa8a6a8a7ccab7ec25559f2eb3401d10626d4fdc3c1442f5",
    "created_at": "1705586782",
    "asset": "lbtc",
    "type": "swap-out",
    "role": "sender",
    "state": "State_SwapCanceled",
    "initiator_node_id": "037f66e84e38fc2787d578599dfe1fcb7b71f9de4fb1e453c5ab85c05f5ce8c2e3",
    "peer_node_id": "023e24602891c28a7872ea1ad5c1bb41abe4206ae1599bb981e3278a121e7895d6",
    "amount": "2000000",
    "channel_id": "812324:2395:1",
    "opening_tx_id": "",
    "claim_tx_id": "",
    "cancel_message": "the prepayment probe was unsuccessful: TemporaryChannelFailure(update=(*lnwire.ChannelUpdate)(0xc005d9f600)({\n Signature: [...removed...],\n ShortChannelID: (lnwire.ShortChannelID) 812324:2395:1,\n Timestamp: (uint32) 1705521406,\n MessageFlags: (lnwire.ChanUpdateMsgFlags) 00000001,\n ChannelFlags: (lnwire.ChanUpdateChanFlags) 00000001,\n TimeLockDelta: (uint16) 142,\n HtlcMinimumMsat: (lnwire.MilliSatoshi) 10000 mSAT,\n BaseFee: (uint32) 1000,\n FeeRate: (uint32) 0,\n HtlcMaximumMsat: (lnwire.MilliSatoshi) 9603000000 mSAT,\n ExtraOpaqueData: (lnwire.ExtraOpaqueData) {\n }\n})\n)",
    "lnd_chan_id": "893159683678470145"
  }
}
```

On the _peer side_ from @jvxis, peerswapd still awaits my payment 
```
pscli listactiveswaps{
  "swaps": [    {
[...]

PR #293, opened 2024-04-02
swap: add a recovery process for AwaitFeeInvoice

YusukeShimizu:
`AwaitFeeInvoice`state to hang in case of state mismatch with peer. https://github.com/ElementsProject/peerswap/issues/290

To eliminate the persistence of uncompleted swaps, execute "cancel" if recovery by restart is executed in `AwaitFeeInvoice` state.


nepet (approved):
ack

PR #294, opened 2024-04-03
lnd: exponential backoff for ProbePayment

YusukeShimizu:
Implement retry mechanism with
exponential backoff for ProbePayment.

Add backoff package to handle retries with exponential backoff. In cln, the default retry time is set to 60 seconds, so the backoff setting is not made.


nepet (approved):
ack

PR #298, opened 2024-06-05
config: fix cln listconfig output

grubles:
CLN PeerSwap `listconfig` currently outputs a `liquid.disabled` field, which doesn't match how `liquidswaps` is set in the `peerswap.conf` file. The output will print as the opposite of how `liquidswaps` is set. This should fix it to properly match how the config option is set.

### DIFF
--- a/clightning/clightning.go
+++ b/clightning/clightning.go
@@ -356,7 +356,7 @@ func (cl *ClightningClient) SetPeerswapConfig(config *Config) {
 		LiquidRpcHost:          config.Liquid.RpcHost,
 		LiquidRpcPort:          config.Liquid.RpcPort,
 		LiquidRpcWallet:        config.Liquid.RpcWallet,
-		LiquidDisabled:         *config.Liquid.LiquidSwaps,
+		LiquidEnabled:          *config.Liquid.LiquidSwaps,
 		PeerswapDir:            config.PeerswapDir,
 	}
 }
--- a/clightning/clightning.go
+++ b/clightning/clightning.go
@@ -356,7 +356,7 @@ func (cl *ClightningClient) SetPeerswapConfig(config *Config) {
 		LiquidRpcHost:          config.Liquid.RpcHost,
 		LiquidRpcPort:          config.Liquid.RpcPort,
 		LiquidRpcWallet:        config.Liquid.RpcWallet,
-		LiquidEnabled:          *config.Liquid.LiquidSwaps,
+		LiquidSwaps:          *config.Liquid.LiquidSwaps,
 		PeerswapDir:            config.PeerswapDir,
 	}
 }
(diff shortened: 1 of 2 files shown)


clightning/clightning.go
```diff
@@ -356,7 +356,7 @@ func (cl *ClightningClient) SetPeerswapConfig(config *Config) {
 		LiquidRpcHost:          config.Liquid.RpcHost,
 		LiquidRpcPort:          config.Liquid.RpcPort,
 		LiquidRpcWallet:        config.Liquid.RpcWallet,
-		LiquidDisabled:         *config.Liquid.LiquidSwaps,
+		LiquidEnabled:          *config.Liquid.LiquidSwaps,
```
YusukeShimizu, line 359: [imo]How about the same name?
```suggestion
		LiquidSwaps:          *config.Liquid.LiquidSwaps,
```

clightning/options.go
```diff
@@ -10,7 +10,7 @@ const (
 	liquidRpcUserOption             = "peerswap-elementsd-rpcuser"
 	liquidRpcPasswordOption         = "peerswap-elementsd-rpcpassword"
 	liquidRpcPasswordFilepathOption = "peerswap-elementsd-rpcpasswordfile"
-	liquidDisabledOption            = "peerswap-elementsd-disabled"
+	liquidEnabledOption             = "peerswap-elementsd-enabled"
```
YusukeShimizu, line 13: Same background.
How about the same name?
```suggestion
	liquidSwapsOption             = "peerswap-elementsd-swaps"
```

clightning/options.go
```diff
@@ -53,7 +53,7 @@ type PeerswapClightningConfig struct {
 	LiquidRpcHost         string `json:"liquid.rpchost"`
 	LiquidRpcPort         uint   `json:"liquid.rpcport"`
 	LiquidRpcWallet       string `json:"liquid.rpcwallet"`
-	LiquidDisabled        bool   `json:"liquid.disabled"`
+	LiquidEnabled         bool   `json:"liquid.enabled"`
```
YusukeShimizu, line 56: ditto
```suggestion
	LiquidSwaps         bool   `json:"liquid.swaps"`
```

YusukeShimizu (approved):
It does not appear that anything will be broken by this change.

The `legacyOptions` in `clightning/options.go` are `deprecated` and may be deleted separately

Also, I will respond to the missing config console print setting in the lwk response.

grubles:
Should we go ahead and remove the `legacyOptions` in this PR? What about `RegisterOptions()`?

grubles:
I went ahead and just renamed the rest. We can open another PR to remove that stuff in the future.

YusukeShimizu:
Yes, I agree.
 I will create it separately.

Issue #299, opened 2024-06-17
`peerswap-listpeers` missing `channels` field due to CLN API removal


grubles:
CLN v24.05 [removed](https://github.com/ElementsProject/lightning/blob/master/CHANGELOG.md#removed) the `channels` field from `listpeers`. This means the `peerswap-listpeers` command is also now missing that field. Reported by @Impa10r on Discord. 

How difficult will it be to use CLN's `listpeerchannels` instead?

Related to #194.

Impa10r:
> CLN v24.05 [removed](https://github.com/ElementsProject/lightning/blob/master/CHANGELOG.md#removed) the `channels` field from `listpeers`. This means the `peerswap-listpeers` command is also now missing that field. Reported by @Impa10r on Discord.
> 
> How difficult will it be to use CLN's `listpeerchannels` instead?
> 
> Related to #194.

Would need to add that method here first:
https://github.com/ElementsProject/glightning/blob/55240d61b52a059ece7d3eca0ed05fe3c413ee54/glightning/lightning.go#L2583

PR #300, opened 2024-07-07
clightning: use CLN's listpeerchannels instead

YusukeShimizu:
The channels field was removed from listpeers in CLN v24.05. 
This means that the peerswap-listpeers command also does not have this field.

Ensure that listchannels are used so that they are still visible in CLN v24.05 and later.
This is a backwards-compatible change that will continue to be displayed in CLN v24.05 and earlier.

Confirmed to work with v24.05 and v23.11.

Fixes https://github.com/ElementsProject/peerswap/issues/299.

### DIFF
--- a/clightning/clightning_commands.go
+++ b/clightning/clightning_commands.go
@@ -647,9 +647,12 @@ func (l *ListPeers) Call() (jrpc2.Result, error) {
 				},
 				PaidFee: paidFees,
 			}
-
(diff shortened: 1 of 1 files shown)


grubles:
Tested ACK dd78a6.

nepet (approved):
ack dd78a61893a2385c7f31cef86667422e699d147a

PR #306, opened 2024-08-19
clightning: liquid testnet port update

YusukeShimizu:
The default port for the liquid testnet in elementsd has been changed from 18891 to 7039.

If the port is not specified, an error like `Could not connect to elements: Post "http://127.0.0.1:18891/wallet/\": EOF` will occur.

Change the default port for peerswap from 18891 to 7039 as well.


grubles:
ACK bce1ae4

PR #310, opened 2024-09-08
txwatcher: improve logging in csvPassedCallback

YusukeShimizu:
Enhances logging in the csvPassedCallback function to provide more context when errors occur.

It now includes the swap ID, transaction ID, and starting block height along with the error message.
This change aids in debugging and troubleshooting issues related to CSV transaction handling.


grubles:
ACK 0ff7ed6

Issue #305, opened 2024-08-04
Please remove [INFO] [MsgListener]: Handler failed: message type not in range


Impa10r:
There are custom messages on LN outside of peerswap's range and peerswap should not spam its log when it sees them.

Thanks!

PR #315, opened 2024-09-16
ignore non-peerswap related custom messages

YusukeShimizu:
Ignore custom messages that are not relevant to peerswap.  
Additionally, it refactors the existing message conversion logic for better clarity and maintainability.
Fixes https://github.com/ElementsProject/peerswap/issues/305


grubles:
I've been running this PR for a few days now and have not seen any `message type not in range` errors when previously there would be a couple every day or so. 

ACK 394cd06.

PR #316, opened 2024-09-18
swap: exclude non-active swap in csvPassedCallback

YusukeShimizu:
csv wait list is always added in `AwaitPaymentOrCsvAction` state but removed only after csvPassedCallback.

`csvPassedCallback` always assumes an active swap, but in normal flows like preimage path, the swap might be complete,
leaving no active swap at this point.
This was not considered before, causing continuous errors.

Previously, if a swap was completed through means other than the `CSV path`, an error message `csv passed callback err: swap does not exist.` would occur unless the peerswap daemon was restarted.  
I have confirmed that this issue no longer occurs.


grubles:
Tested ACK 2132b1e

Issue #320, opened 2024-10-01
peerswap can't read CLN >24.08 listconfig output


grubles:
In CLN >24.08, the `listconfig` structure has changed causing peerswap to misread the `allow-deprecated-apis` field as `false`.

When that field is set to false, peerswap exits and print a log message explaining why. Since peerswap can't currently read the new API structure, it will always exit. We should update how peerswap (or more appropriately glightning?) reads CLN's `listconfig`. 

"Instead of direct members with names equal the config variable, there's now a `configs` sub-object containing a member with details of each config setting[^1]"

[^1]: https://docs.corelightning.org/docs/deprecations

PR #321, opened 2024-10-03
clightning: compatibility issue with v24.11

YusukeShimizu:
Resolves a compatibility issue with clightning.
v24.11 and later versions where the `allow-deprecated-apis` option may be nested within a configs object of ListConfigs.

Now correctly parses the configuration to determine if deprecated APIs are allowed, ensuring compatibility with both older and newer clightning versions.

Fixes: https://github.com/ElementsProject/peerswap/issues/320

### DIFF
--- a/clightning/config.go
+++ b/clightning/config.go
@@ -99,6 +99,28 @@ func SetPeerswapPaths(plugin *glightning.Plugin) Processor {
 	}
 }
 
+type AllowDeprecatedAPIs struct {
+	ValueBool bool   `json:"value_bool"`
+	Source    string `json:"source"`
+}
+type Configs struct {
+	AllowDeprecatedAPIs *AllowDeprecatedAPIs `json:"allow-deprecated-apis,omitempty"`
+}
+type ClnConfig struct {
+	AllowDeprecatedApis bool     `json:"allow-deprecated-apis,omitempty"`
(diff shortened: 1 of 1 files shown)


grubles:
The peerswap plugin now starts when running with CLN master but prints this new error. Is another new API deprecation/removal causing it? 
```
2024-10-03T15:17:52.764Z INFO    plugin-peerswap: peerswap initialized
2024-10-03T15:17:53.434Z DEBUG   lightningd: channel_gossip: no longer in startup mode
2024-10-03T15:17:53.460Z INFO    lightningd: JSON COMMAND sendcustommsg: Invalid parameter node_id (should be a node id): token '""'
2024-10-03T15:17:53.462Z DEBUG   plugin-peerswap: poll_service: could not send msg to : -32602:node_id: should be a node id: invalid token (see logs for details)
```

YusukeShimizu:
The error seems to be occurring when sending a custom message for polling.  
It's unclear whether this is due to deprecation or removal.

I don't think the failure of the poll will result in an immediate impact.

Once a new version of cln is released on nix, I plan to run integration tests to verify.

grubles:
Ok I tested a swap using v24.08.1 and master. The `poll_service` doesn't seem to break swaps AFAICT.

 ACK 663acfd

Issue #325, opened 2024-10-29
Swaps canceled due to "missing transaction" when using LWK backend


grubles:
When a peer that is running LWK tries to broadcast a swap transaction, sometimes it will fail and the cancel message will be `missing transaction`.

Example logs of this when running LND + LWK + peerswapd:

```
024/10/29 17:35:42 [DEBUG] [FSM] event:id: 3f9a179cec432d04bc4442d3c5a6b9b95311d99181bafd1a65bc7eb0c1d3aa8b, Event_SwapInSender_OnSwapInRequested on                                         
2024/10/29 17:35:42 [INFO] [Swap:3f9a179cec432d04bc4442d3c5a6b9b95311d99181bafd1a65bc7eb0c1d3aa8b] Start new swap-in: peer: 036e680823a129ac9a2eeb8c43514a5b6cb09ae28e3e698a0e34aaa8865d6cc344
 chanId: 194668:21:0 initiator: 035ca2fe4793a5e789ce846062eb4834f573c060d9200ce77544a29b48a0aa5923 amount 100000001                                                                           
2024/10/29 17:35:42 [DEBUG] [FSM] event:id: 3f9a179cec432d04bc4442d3c5a6b9b95311d99181bafd1a65bc7eb0c1d3aa8b, Event_ActionSucceeded on State_SwapInSender_CreateSwap                          
2024/10/29 17:35:42 [DEBUG] [FSM] event:id: 3f9a179cec432d04bc4442d3c5a6b9b95311d99181bafd1a65bc7eb0c1d3aa8b, Event_ActionSucceeded on State_SwapInSender_SendRequest                         
2024/10/29 17:35:43 [DEBUG] [Messenger] From: 036e680823a129ac9a2eeb8c43514a5b6cb09ae28e3e698a0e34aaa8865d6cc344 got msgtype: a459 for swap: 3f9a179cec432d04bc4442d3c5a6b9b95311d99181bafd1a6
5bc7eb0c1d3aa8b                                                                                                                                                                               
2024/10/29 17:35:43 [DEBUG] [FSM] event:id: 3f9a179cec432d04bc4442d3c5a6b9b95311d99181bafd1a65bc7eb0c1d3aa8b, Event_SwapInSender_OnAgreementReceived on State_SwapInSender_AwaitAgreement
2024/10/29 17:35:43 [DEBUG] using minimum fee rate of 0.1 sat/vbyte                                                                                                                           
[...]

PR #326, opened 2024-10-31
lwk: add backoff for send

YusukeShimizu:
Handles retries  in case of a "missing transaction" error.  
It uses an exponential backoff strategy for retries, with a maximum of 5 retries.  

fixes: https://github.com/ElementsProject/peerswap/issues/325

### DIFF
--- a/electrum/client.go
+++ b/electrum/client.go
@@ -3,7 +3,11 @@ package electrum
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
+	"strings"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/checksum0/go-electrum/electrum"
 	"github.com/elementsproject/peerswap/log"
 )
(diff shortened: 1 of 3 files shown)


grubles:
Ok I think I've triggered the backoff for this swap attempt:
```
$ ./builds/peerswap/out/pscli swapin --sat_amt 100000001 --channel_id 214039729557274624 --asset lbtc
2024/11/02 09:26:28 rpc error: code = Unknown desc = failed to fund transaction: permanent error: failed to call api request: POST http://localhost:32111 giving up after 1 attempt(s): context deadline exceeded
```

Though in other canceled attempts I will get this error:

```
2024/11/01 12:57:11 [DEBUG] [FSM] event:id: 4ee04c58b8a037004b487211584102504413dc6de06e933cf17511b2a7a12eab, Event_SwapInSender_OnSwapInRequested on 
2024/11/01 12:57:11 [INFO] [Swap:4ee04c58b8a037004b487211584102504413dc6de06e933cf17511b2a7a12eab] Start new swap-in: peer: 036e680823a129ac9a2eeb8c43514a5b6cb09ae28e3e698a0e34aaa8865d6cc344 chanId: 194668:21:0 initiator: 035ca2fe4793a5e789ce846062eb4834f573c060d9200ce77544a29b48a0aa5923 amount 100000001
2024/11/01 12:57:11 [DEBUG] [FSM] event:id: 4ee04c58b8a037004b487211584102504413dc6de06e933cf17511b2a7a12eab, Event_ActionSucceeded on State_SwapInSender_CreateSwap
2024/11/01 12:57:11 [DEBUG] [FSM] event:id: 4ee04c58b8a037004b487211584102504413dc6de06e933cf17511b2a7a12eab, Event_ActionSucceeded on State_SwapInSender_SendRequest
2024/11/01 12:57:12 [DEBUG] [Messenger] From: 036e680823a129ac9a2eeb8c43514a5b6cb09ae28e3e698a0e34aaa8865d6cc344 got msgtype: a459 for swap: 4ee04c58b8a037004b487211584102504413dc6de06e933cf17511b2a7a12eab
2024/11/01 12:57:12 [DEBUG] [FSM] event:id: 4ee04c58b8a037004b487211584102504413dc6de06e933cf17511b2a7a12eab, Event_SwapInSender_OnAgreementReceived on State_SwapInSender_AwaitAgreement
2024/11/01 12:57:13 [INFO] error getting fee: cannot estimate fee for 7 blocks.
2024/11/01 12:57:13 [DEBUG] using minimum fee rate of 0.1 sat/vbyte
2024/11/01 12:57:13 [INFO] [FSM] Action failure failed to get raw transaction: missing transaction
2024/11/01 12:57:13 [DEBUG] [FSM] event:id: 4ee04c58b8a037004b487211584102504413dc6de06e933cf17511b2a7a12eab, Event_ActionFailed on State_SwapInSender_BroadcastOpeningTx
2024/11/01 12:57:13 [DEBUG] [FSM] Canceling because of failed to get raw transaction: missing transaction
2024/11/01 12:57:13 [DEBUG] [FSM] event:id: 4ee04c58b8a037004b487211584102504413dc6de06e933cf17511b2a7a12eab, Event_ActionSucceeded on State_SendCancel
2024/11/01 12:57:13 [INFO] [Swap:4ee04c58b8a037004b487211584102504413dc6de06e933cf17511b2a7a12eab] Swap canceled. Reason: failed to get raw transaction: missing transaction
```

YusukeShimizu:
Thank you for your investigation.

The core issue lies in the failure to retrieve the result after broadcasting the transaction. In this situation, even though the transaction itself is valid on the network, the system cancels the swap, causing a problem where the balance decreases and cannot be recovered.

- **Misunderstanding**: Initially, I assumed that a `missing transaction error` occurred during the `fund transaction` (LWK's `send` command).
- **Reality**: The `fund transaction` itself is executed correctly. The issue arises after broadcasting the transaction, where it fails to retrieve the result. Due to this failure, while the swap is canceled, the transaction has already been processed, leading to a decrease in balance that cannot be reclaimed.
## Solution
I will introduce backoff handling when retrieving transaction results. This can potentially mitigate errors caused by temporary failures or delays.
## Considerations
The case where a swap is canceled even though an opening tx was broadcast may not only occur with LWK/electrs but could also happen with other backends. However, the probability is low.

In the future, it may be necessary to implement a manual fund recovery command. For now, we will focus on stabilizing within the LWK environment and consider further actions if issues arise.
### Additional Information
"failed to fund transaction: permanent error" occurs when there’s an error with LWK's send command but does not trigger backoff. Backoff only occurs when a "missing transaction" error is returned.

PR #327, opened 2024-11-08
lwk: increase default context timeout

YusukeShimizu:
increase default context timeout from 5 seconds to 20 seconds.


grubles:
Running `lbtc-getbalance` took 7.2 seconds this time, which would've produced an error previously but with the new timeout was given enough time to complete. Tested ACK [79274df](https://github.com/ElementsProject/peerswap/pull/327/commits/79274dfada887d852eeeac1477562227fd2ca588)
```
$ time ./builds/peerswap/out/pscli lbtc-getbalance
{
  "sat_amount": "93373298"
}

real    0m7.204s
user    0m0.011s
sys     0m0.010s
```

PR #328, opened 2024-11-08
golangci-lint configuration

YusukeShimizu:
Reduce the timeout for golangci-lint from 5 minutes to 1 minute.
Set issues-exit-code to 1 to ensure lint issues cause a non-zero exit.
Enable tests to be run during the linting process.
Use default directories for skipping.
Allow parallel runners to improve linting performance.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -101,4 +101,8 @@ issues:
         - dupl
         - gosec
 run:
-  timeout: 5m
+  timeout: 1m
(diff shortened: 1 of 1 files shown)


grubles:
ACK [50f0d43](https://github.com/ElementsProject/peerswap/pull/328/commits/50f0d43b43ce38fc7841f6382b7f813bba4ed1fd)

PR #318, opened 2024-09-23
clightning: add shutdown hook for plugin

YusukeShimizu:
Introduces a shutdown hook to the clightning plugin, ensuring that the plugin is gracefully stopped when c-lightning shuts down.

* Subscribe to the 'shutdown' event.
* Updated glightning dependency t o include the 'shutdown' event. https://github.com/ElementsProject/glightning/pull/20

fixes: https://github.com/ElementsProject/peerswap/issues/307

PR #331, opened 2024-11-20
mod: glightning to latest version

YusukeShimizu:
- Update the glightning dependency in go.mod.


YusukeShimizu:
This is an update to lock the version that has been verified in https://github.com/ElementsProject/peerswap/pull/318.  
Since there are no behavioral changes, I will merge it.

PR #348, opened 2025-01-31
multi: acceptdiscountctIsEnabled check in setup

YusukeShimizu:
In the process of launching peerswap, when elementsd is used, a check for `acceptdiscountctIsEnabled` is added.  
The `acceptdiscountctIsEnabled` checks whether the acceptdiscountct feature is enabled by decoding a hard-coded transaction example and verifying if the decoded transaction contains the DiscountVirtualSize field.

Depends on https://github.com/ElementsProject/glightning/pull/22


grubles:
Tested ACK [bbfe93e](https://github.com/ElementsProject/peerswap/commit/bbfe93efe05f75718d31aeba1ab2c2f0880d31f3)

PR #349, opened 2025-02-05
deps: glightning version reference

YusukeShimizu:
Replace old glightning commit with the new version Reflect changes in go.mod and go.sum


YusukeShimizu:
This is the incorporation from glightning's master.  
Since there's no change in behavior, I will merge it.

PR #297, opened 2024-05-29
mod: update package lnd v0.18.0-beta.rc3

YusukeShimizu:
Updated github.com/lightningnetwork/lnd package to v0.18.4-beta.rc1. 
Also updated golang version to 1.22 accordingly.

> [!IMPORTANT]  
We plan to merge after the stable release of `v4.0`. This is crucial information necessary for users to succeed.

PR #350, opened 2025-02-06
multi: migrate protoc to buf

### LINKED ISSUE #343: Introduce Buf for Proto Generation
Our current https://github.com/ElementsProject/peerswap/tree/9dc5c5175d59eabd611c993b09c97fd151f1acf1/peerswaprpc relies on `protoc` for generating code from `.proto` files. While `protoc` has served its purpose,it has several challenges:

1. **Dependency Management**: Maintaining imports and managing shared proto files is becoming cumbersome.
2. **Tooling Limitations**: Limited support for modern workflows, such as linting, breaking change detection, and streamlined plugin usage.

I would like to properly perform validation on proto, and for that, I believe [protovalidate](https://github.com/bufbuild/protovalidate) is appropriate. First, it is necessary to migrate to [Buf](https://buf.build).  
Buf offers a more robust and developer-friendly solution for managing proto files and generating code.

shelter2759:
Issue number: resolves #343 

---------

## What is the current behavior?

- In the peerswaprpc folder, the code is generated by protoc.
## What is the new behavior?

- Introduce buf and migrate from protoc.
- Modify devcontainer.json to install buf.
- Create buf.yaml and buf.gen.yaml.
- Modify makefile and Dockerfile that use protoc command to buf command.
- Delete unnecessary gen_protos.sh files
- Generate code in peerswaprpc folder using buf. 

## Other information
- [Buf migrate-from-protoc](https://buf.build/docs/migration-guides/migrate-from-protoc/)


shelter2759:
@YusukeShimizu 
Hi,
I would like to ask a question about something I noticed while dealing with this issue.

After installing buf package with go, I just ran "go mod tidy -e." command.
I saw the output as shown in the screenshot.
image

Does this one need to be fixed?
If so, I would like to use a different PR to handle it.

Here is some information that may be referenced.
https://github.com/grpc/grpc-go/issues/6264

Thanks.

YusukeShimizu:
Thank you for the PR!  
I really appreciate your effort in catching up with the insufficient documentation and helping out.  
I ran the CI tests and it seems the build is failing, so could you please take a look?

Up to now, we’ve been able to generate the proto files with the following commands:
```sh
cd peerswaprpc
make all-rpc-docker
```

For this PR, how do you plan to generate the proto files? If possible, it would be great if you could add instructions to the documentation.

From what I’ve observed, the build failure appears to be due to a mismatch between the version of the generated gRPC code and the gRPC library we’re using. I see two potential approaches to address this:

## Approach 1: Update the library versions
Updating the Go dependencies (e.g. google.golang.org/grpc) to v1.64.0 or later should align with the generated code, thus resolving the error.  
(Reference: https://github.com/ElementsProject/peerswap/pull/297 addresses dependency updates, so possibly incorporating that would fix the issue.)

## Approach 2: Pin the version to match the existing environment
This involves fixing the version of the buf plugin. By using the existing `make all-rpc-docker` process to install a specific version of the protoc-gen tool, we should be able to generate proto files compatible with the current setup.  
In this PR, it might help to first generate code with the fixed version (matching the existing environment) to ensure the build passes.

shelter2759:
@YusukeShimizu 
Thank you for suggesting the approach.
I have adopted Approach 2 that you proposed and modified the Makefile and Dockerfile to generate files using buf with the same commands as the existing ones.
https://github.com/ElementsProject/peerswap/commit/0bd9afaf2a0d1116aa38c7772f5fcd123558bc95

I want to run build tests in GitHub Actions. Do I need to wait until it runs automatically or  is there any other way to run it?

I’m not yet familiar with the rules of this project, so I apologize for asking so many beginner questions. I don’t mean to cause trouble, but I would greatly appreciate your response.

YusukeShimizu:
In this repository, `the Approval for running fork pull request workflows from contributors setting` is configured as `Require approval for first-time contributors. Only users who have never had a commit or pull request merged into this repository will require approval to run workflows.`.

It appears that the action was not triggered for commit [0bd9afa ](https://github.com/ElementsProject/peerswap/pull/350/commits/0bd9afaf2a0d1116aa38c7772f5fcd123558bc95). The cause is unclear, but could you please resolve the conflicts and push again?

shelter2759:
@YusukeShimizu 
Apologies for the force push.
I have reverted the deletion of gen_proto.sh, which was the cause of the conflict, and kept the file.

Thank you for your understanding. Please let me know if there’s anything else I should address.

YusukeShimizu (approved):

YusukeShimizu:
Thank you for not only migrating to buf but also thoughtfully considering the .devcontainer setup.  
It seems stable at this point, and I think that’s great.  

In the future, if we can incorporate buf into ‘nix develop’ and reduce our reliance on Docker, we can further lower the barrier to setting things up.

YusukeShimizu:
This change is an update to the development experience and is not expected to affect the behavior, so I will merge it.

PR #354, opened 2025-02-11
multi: Modify buf.yaml for Buf lint and Format with Buf CLI Comamnd("buf format -w")

shelter2759:
Issue number: none but related #343 

---------

## What is the current behavior?

- Buf lint configuration in buf.yaml does not exist.
- proto file is not formatted with Buf CLI Command(buf format -w)

## What is the new behavior?

- Buf lint settings in buf.yaml(lint settings are set to ignore errors in the current source code.)
- Execute “buf format -w” and update the proto file (no change in content)
- small changes in Dockerfile
- Add Github Action workflow to execute buf cli (lint, format, breaking)

## Does this introduce a breaking change?
- [x] No

## Other information
- https://buf.build/docs/ci-cd/github-actions/


YusukeShimizu:
Thank you for continuing to provide changes aimed at improving the development experience.  
How about updating so that developers can also run `buf format -w`, `buf lint`, and `buf breaking`, and furthermore adding these steps to GitHub Actions so that any problematic changes are blocked by CI checks?  
I acknowledge the changes of buf.yaml.

shelter2759:
@YusukeShimizu 
I added job to run buf cli(buf-action) in github action workflow.(268c6b9e5b9ce13c7efb2e4206c1d9fe91f2072b)
- buf-action (https://github.com/marketplace/actions/buf-action).

Please check when you have time.
Thank you.

YusukeShimizu:
[IMO] I think introducing buf-action is appropriate, and there’s no issue with the CI content.  
I’d like to confirm one point: it’s not entirely clear how developers are expected to run Buf commands (e.g., buf format -w) locally. We can generate code with “make all-rpc-docker,” but what should we do for other commands?

If you’re in the dev container, Buf is installed using “postCreateCommand,” but what happens in other cases? If there isn’t a provided method, I’d like to document that you either use Docker (like “make all-rpc-docker”) or individually install Buf yourself.

YusukeShimizu:
It seems that [all-rpc](https://github.com/ElementsProject/peerswap/blob/master/peerswaprpc/Makefile#L9) is also a command that requires prior environment setup.  
In line with the above considerations, it may be necessary to either update or remove it

YusukeShimizu (approved):
LGTM.  
We might consider installing buf via nix flake, but it’s already quite user-friendly as it is. Thanks!  

Since this change is intended to enhance the developer experience without affecting behavior, I’ll merge it.

Issue #355, opened 2025-02-13
Incorrect "exceeding receivable amount_msat" error


Impa10r:
v4.0rc1 on testnet3, the error should not be triggered:

```
$ pscli swapin --sat_amt=507914 --channel_id=2796274673227464705 --asset=lbtc
2025/02/13 16:57:47 rpc error: code = Unknown desc = exceeding receivable amount_msat: 507914000
peerswap $ pscli listpeers
{
  "peers":  [
    {
      "node_id":  "020ed4ea09c56c3e398190ae5ac60504fe4e56d6989d49e2f7845f14608c546854",
      "swaps_allowed":  true,
      "supported_assets":  [
        "btc",
        "lbtc"
      ],
      "channels":  [
        {
          "channel_id":  "2796274673227464705",
          "local_balance":  "481141",
          "remote_balance":  "517914",
          "active":  true
        }
      ],
```

--sat_amt=507913 works, so the cushion required is 10,001. 

However, on a mainnet I saw the cushion required 100,001. What is the exact formula?

YusukeShimizu:
[Referring to the code](https://github.com/ElementsProject/peerswap/blob/master/lnd/client.go#L147), the basis for calculating “ReceivableMsat” is as follows:
```go
receivable := (remoteBalanceSat - remoteChanReserveSat) * 1000
if maxHtlcAmtMsat != 0 {
    receivable = min(receivable, maxHtlcAmtMsat)
}
```
In other words, “ReceivableMsat” is the smaller of (remoteBalance minus remoteConstraints().ChanReserveSat) and maxHtlcAmtMsat. Might it be possible that this issue arises because the channel reserve differs between testnet and mainnet?

Impa10r:
Will have a look, thanks. The incorrect fact was that the swap amount 507914 did not exceed amount_msat: 507914000 (they are equal).

PR #356, opened 2025-02-14
Fix "exceeding amount_msat" check

Impa10r:
https://github.com/ElementsProject/peerswap/issues/355

Replace <= with <

### DIFF
--- a/swap/actions.go
+++ b/swap/actions.go
@@ -602,7 +602,7 @@ func (r *PayFeeInvoiceAction) Execute(services *SwapServices, swap *SwapData) Ev
 	requiredBalance := swap.SwapOutRequest.Amount*1000 + msatAmt
 
 	// Check if the spendable balance (sp) is sufficient
-	if sp <= requiredBalance {
+	if sp < requiredBalance {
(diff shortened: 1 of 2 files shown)


YusukeShimizu:
Thank you for the revisions.  
I believe it’s correct to allow for situations where sp (spendable amount) and rs (receivable amount) are exactly the same. 
 
 However, the spendable amount is also checked again when executing PayFeeInvoiceAction. In order to implement the intended changes, wouldn’t the modification also be necessary?
 https://github.com/ElementsProject/peerswap/blob/178522abe397fe8b48acc01b7b3035413930fc36/swap/actions.go#L654

Impa10r:
> However, the spendable amount is also checked again when executing PayFeeInvoiceAction. In order to implement the intended changes, wouldn’t the modification also be necessary?
> 
> https://github.com/ElementsProject/peerswap/blob/178522abe397fe8b48acc01b7b3035413930fc36/swap/actions.go#L654

agree, did not notice

YusukeShimizu (approved):
thank you!

PR #369, opened 2025-04-03
Add testnet4 support

Impa10r:
lnd and cln now support testnet4 https://github.com/lightningnetwork/lnd/pull/9620, ps should too


go.mod
```diff
@@ -8,7 +8,7 @@ toolchain go1.23.4
 
 require (
 	github.com/benbjohnson/clock v1.3.0 // indirect
-	github.com/btcsuite/btcd v0.24.3-0.20240921052913-67b8efd3ba53
```
YusukeShimizu, line 11: In lnd as well, rather than using a fixed version, it references a commit in the same way.

YusukeShimizu (commented):
It appears that the same consideration is necessary for `validateNetwork` as well.
https://github.com/ElementsProject/peerswap/blob/master/swap/messages.go#L123

Could you please check if the swap is functioning properly?  

Also, could you handle the corresponding change on the CLN side? It seems that support for testnet4 was recently added https://github.com/ElementsProject/lightning/pull/8171.

Impa10r:
thanks. updated and confirm working with lnd. struggling to test with cln so far.

go.mod
YusukeShimizu: Ah, this is just a note for my own record. Please disregard.

YusukeShimizu:
Would you mind reverting that merge commit and rebasing instead?

YusukeShimizu (approved):
LGTM.  
The changes are limited to non-production areas (tests), so there’s no impact on runtime behavior or mainnet-facing code. All checks pass, so I’m approving.

PR #264, opened 2023-12-23
Premium

YusukeShimizu:
This PR adds premium to peerswap.
[design doc](https://docs.google.com/document/d/1enPKthscJCRX_bG_12fwcUjfm9-FdTBoRvPIL1HLR1k/edit#heading=h.uoxht6u5sf15)

### doing swap
A premium limit will be added to the existing swap command.
If not set, default is 0.

#### cln
```bash
lightning-cli peerswap-swap-out [short channel id] [amount in sats] [asset: btc or lbtc] [premium limit in sats]
```

#### lnd
```bash
pscli swapout --sat_amt=[sat amount] --channel_id=[chan_id from above] --asset=btc --premium_limit=[premium limit]
```

### configure premium rate
The premium rate is pert per million and can be set for swapin and swapout respectively.
`swap_amount * premium_rate_ppm / 1000000` is the premium.
Negative values can also be set to give a discount

For usage instructions, please refer to [YusukeShimizu/peerswap@premium-feature/docs/usage.md#premium](https://github.com/YusukeShimizu/peerswap/blob/premium-feature/docs/usage.md#premium).

The gRPC interface can be found here: [YusukeShimizu/peerswap@premium-feature/peerswaprpc/peerswaprpc.proto#L35-L42](https://github.com/YusukeShimizu/peerswap/blob/premium-feature/peerswaprpc/peerswaprpc.proto#L35-L42).

### examples
If the max_acceptable_premium is lower than responder.
```sh
./bin/pscli peerswap1 swapout --channel_id $CHANID --sat_amt 1000000 --asset btc --premium_limit 1
2023/12/23 02:02:36 rpc error: code = Unknown desc = from the LND peer: unacceptable premium: 100, limit: 1
```

success
```sh
./bin/pscli peerswap1 swapout --channel_id $CHANID --sat_amt 1000000 --asset btc --premium_limit 100
{
  "swap": {
    "id": "eb5f41a515f737956300a38ab1f5286a67d580c4d3f838afdf48537c1133ed69",
    "created_at": "1702532654",
    "asset": "btc",
    "type": "swap-out",
    "role": "sender",
    "state": "State_SwapOutSender_AwaitTxConfirmation",
    "initiator_node_id": "02ea1fd791bb1b639e7119ac54505e3f553a7bb942877ccbb359affa8547046988",
    "peer_node_id": "034066fecfcdad7cb2f002c8c4234dbab26688db9617509ede4c7372c79ab94d0d",
    "amount": "1000000",
[...]

PR #371, opened 2025-04-06
rpc: return raw peerswaprpc.PremiumRate pointers

YusukeShimizu:
Fix returning proto definitions to ease unmarshal for PremiumRate.
We previously format peerswaprpc.PremiumRate for readability.  https://github.com/ElementsProject/peerswap/pull/264#issuecomment-2730250057

But, that makes unmarshal over RPC calls inconvenient.
To callable, return raw proto definitions directly to fix RPC usage issues.

Some 0-value fields may be omitted due to proto generation limits.
Removing omit empty tags later may resolve those readability issues.

In cln, CLI output and RPC calls are not separated.  
Because of this, it is not possible to make the CLI output human-readable while returning raw proto for RPC.

### DIFF
--- a/clightning/clightning_commands.go
+++ b/clightning/clightning_commands.go
@@ -3,7 +3,6 @@ package clightning
 import (
 	"crypto/rand"
 	"encoding/hex"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"math/big"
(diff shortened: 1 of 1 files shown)


Impa10r:
Merged into my testnet4 branch and it fixed the issue with CLN premiums. Thanks!

grubles:
tACK [b242fae](https://github.com/ElementsProject/peerswap/pull/371/commits/b242fae25ac00f19a38e1ec226896cad2f03c4c6)

PR #389, opened 2025-07-08
refactor: use the built-in max/min

bytetigers:
Use the built-in [max/min](https://pkg.go.dev/builtin@go1.21.0#max) function to simplify the code.

Inspired by https://github.com/ElementsProject/peerswap/pull/367


YusukeShimizu (approved):
Thank you!

PR #396, opened 2025-08-25
chore: make function comment match function name[skip ci]

pxwanglu:
make function comment match function name


pxwanglu:
friendly ping~

PR #413, opened 2025-12-09
Increase gRPC message size limits for peerswapd server and REST gateway

mrfelton:
The `/swaps` REST endpoint fails with `ResourceExhausted` errors when responses exceed the default 4 MiB gRPC limit. The server and REST gateway weren't configured with increased limits, unlike the client-side callers which already use 200-500 MiB.

### Changes

- Add `maxMsgRecvSize` constant (500 MiB) to match existing limits in `lnd/connect.go` and `testframework/proxy.go`
- Configure gRPC server with `MaxRecvMsgSize` and `MaxSendMsgSize`
- Configure REST gateway dial options with `MaxCallRecvMsgSize`

```go
grpcSrv := grpc.NewServer(
    grpc.MaxRecvMsgSize(maxMsgRecvSize),
    grpc.MaxSendMsgSize(maxMsgRecvSize),
)

opts := []grpc.DialOption{
    grpc.WithTransportCredentials(insecure.NewCredentials()),
    grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(maxMsgRecvSize)),
}
```

### DIFF
--- a/cmd/peerswaplnd/peerswapd/main.go
+++ b/cmd/peerswaplnd/peerswapd/main.go
@@ -51,6 +51,10 @@ import (
 
 const (
 	minLndVersion = float64(14.1)
+	// maxMsgRecvSize is the largest message our gRPC server and REST gateway
+	// will receive. This is set to 500 MiB to be consistent with other
+	// gRPC clients in this repo (e.g. lnd/connect.go, testframework/proxy.go).
+	maxMsgRecvSize = 1 * 1024 * 1024 * 500
 )
 
 var GitCommit string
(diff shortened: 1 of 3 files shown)


YusukeShimizu:
Thanks for addressing this.

Instead of raising the gRPC message size limit, how about adding paging (plus minimal filtering if needed) to ListSwaps / ListPeers so responses are bounded by default, since simply increasing the limit hides the symptom but still leaves response size and memory usage growing monotonically as history accumulates.

For backward compatibility, we could either keep the current behavior when paging params are omitted (return all) or introduce a new RPC.

Issue #301, opened 2024-07-16
[feature request] Zip log file and start a new one when it grows past 10mb


Impa10r:
Currently, there is no limit how big your log file can be. So it becomes slow to open eventually. I suggest you borrow code from LND to archive it after 10mb size and start a new one.

PR #423, opened 2026-01-23
Adding log rotation using lumberjack

sergey3bv:
@YusukeShimizu
@nepet
Addressing #301

### DIFF
--- a/cmd/peerswaplnd/config_test.go
+++ b/cmd/peerswaplnd/config_test.go
@@ -11,6 +11,77 @@ import (
 	"github.com/vulpemventures/go-elements/network"
 )
 
+func TestLogRotationConfigValidate(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name    string
+		cfg     peerswaplnd.LogRotationConfig
+		wantErr bool
+	}{
+		{
+			name: "valid values",
+			cfg: peerswaplnd.LogRotationConfig{
+				MaxSize:    10,
+				MaxBackups: 5,
+				MaxAge:     28,
+				Compress:   true,
+			},
+			wantErr: false,
+		},
+		{
+			name: "maxsize zero",
+			cfg: peerswaplnd.LogRotationConfig{
+				MaxSize:    0,
+				MaxBackups: 5,
+				MaxAge:     28,
+			},
+			wantErr: true,
+		},
+		{
+			name: "maxsize negative",
+			cfg: peerswaplnd.LogRotationConfig{
+				MaxSize:    -1,
+				MaxBackups: 5,
+				MaxAge:     28,
+			},
+			wantErr: true,
+		},
+		{
+			name: "maxbackups negative",
+			cfg: peerswaplnd.LogRotationConfig{
+				MaxSize:    10,
+				MaxBackups: -1,
+				MaxAge:     28,
+			},
+			wantErr: true,
+		},
+		{
+			name: "maxage negative",
+			cfg: peerswaplnd.LogRotationConfig{
+				MaxSize:    10,
+				MaxBackups: 5,
+				MaxAge:     -1,
+			},
+			wantErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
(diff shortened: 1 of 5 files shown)


cmd/peerswaplnd/peerswapd/main.go
```diff
@@ -597,23 +596,26 @@ type LndLogger struct {
...
 }
 
-func NewLndLogger(cfg *peerswaplnd.PeerSwapConfig) (*LndLogger, func() error, error) {
-	logFile, err := os.OpenFile(filepath.Join(cfg.DataDir, "log"), os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
-	if err != nil {
-		return nil, nil, err
-	}
-	w := io.MultiWriter(os.Stdout, logFile)
+func NewLndLogger(cfg *peerswaplnd.PeerSwapConfig) (*LndLogger, error) {
+	logFile := filepath.Join(cfg.DataDir)
```
YusukeShimizu: I may be missing something, but since lumberjack.Logger.Filename is currently set to cfg.DataDir, could this end up using a directory where a file path is expected?  
If that happens, wouldn’t it conflict with the later code that expects cfg.DataDir to remain a directory for swaps and peersync.db?
Would it make sense to set Filename to filepath.Join(cfg.DataDir, "log") and call os.MkdirAll(cfg.DataDir, 0755) before NewLndLogger(cfg)?

YusukeShimizu (changes requested):
Thank you for the log rotation update; this is a valuable improvement overall.  
I have one concern that looks blocking, so I’m marking this as request changes for now and would appreciate a quick confirmation.

cmd/peerswaplnd/peerswapd/main.go
sergey3bv: Great call, indeed Lumberjack waits for file path instead of directory

sergey3bv:
Hey, @YusukeShimizu, I updated the PR accordingly, could you please review it? Also are there any issues I can help with?

YusukeShimizu:
Thanks for addressing the earlier feedback and improving the log rotation overall!

If my understanding is correct, there are two points we might need to address before merging:

1. Existing nodes currently use <datadir>/log as a file, but it looks like this PR changes the same path into a directory. Please correct me if I'm wrong, but wouldn't this cause an in-place upgrade to fail at startup? We might need to add a migration step for the old file or keep the path backward-compatible.

2. It seems logrotation.maxsize, maxbackups, and maxage are being accepted without validation. I'm concerned that invalid values (especially maxsize <= 0) could silently break file logging rather than failing fast at startup.

Let me know your thoughts on this!

sergey3bv:
> 1. Existing nodes currently use /log as a file

As I am thinking about it now we can leave the current flow and store rotated logs nearby. For example, logs will still be written in `log` and rotated ones will be in `rotated_logs` or even in user provided location. This way PR will produce less friction, what do you think?

sergey3bv:
> 2\. I'm concerned that invalid values (especially maxsize <= 0) could silently break file logging

Good point, I will fix it right away

YusukeShimizu:
That sounds like a good direction to me. Keeping the active log at the current <datadir>/log path would avoid the upgrade compatibility issue, and it should reduce friction for existing deployments.

sergey3bv:
Hey, @YusukeShimizu, I have updated the PR accordingly, could you please take a look?

cmd/peerswaplnd/config_test.go
```diff
@@ -11,6 +11,77 @@ import (
...
+				MaxSize:    10,
+				MaxBackups: 5,
+				MaxAge:     -1,
+			},
+			wantErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
YusukeShimizu, line 71: [Nit] `tt := tt` looks unnecessary here

YusukeShimizu (approved):
LGTM, thanks for addressing this.  
The log rotation/compression behavior now matches the goal of issue #301, and the earlier datadir/log path and config validation concerns also look resolved.

Issue #394, opened 2025-07-30
[feature request] sanity checks for premium rates


xmrk-btc:
Please add some sanity checks to disallow setting premium rates which are not in node operator's interest. 
1. Check that SWAP_IN rate is negative or zero, and that SWAP_OUT rate is positive or zero. I saw one my peer doing exactly the opposite, his SWAP_OUT rate is negative, so he is paying people to "steal" his inbound liquidity. (Or maybe I am misinterpreting, but then my rates are wrong and sanity check would help me.)
2. Check that BTC_SWAP_IN_rate + BTC_SWAP_OUT_rate > 0 , likewise for LBTC. This is to prevent peers from profiting by doing SWAP_OUT and then SWAP_IN of same amount. Possible inconvenience here is that if I have to first  increase SWAP_OUT rate and then decrease SWAP_IN rate, doing it in reverse order could be disallowed by sanity check. So include such hint about the order of operations in the error message.
3. Maybe check that BTC_SWAP_IN_rate + LBTC_SWAP_OUT_rate > boltz_rate_for_lbtc_pegout. This is to prevent peers from doing LBTC swap-out, peg-out at boltz, and doing swap-in with obtained BTC. Maybe just use constant like 1000 ppm for boltz_rate_for_lbtc_pegout.
4. Maybe check that LBTC_SWAP_IN_rate + BTC_SWAP_OUT_rate > 0. Prevents peers from doing BTC swap-out, peg-in those BTC (which is free, hence 0), and then LBTC swap-in.

YusukeShimizu:
Thanks for opening the issue. Your point about the risk of mis-configuring the premium rate is well taken, and the concrete examples you provided make the potential impact very clear.

That said, PeerSwap is meant to be a reference implementation of the peerswap protocol, and I feel that adding a hard sanity check at this layer is out of scope. Negative premium rates are allowed by the spec, and some node operators may have perfectly valid reasons to use them—for example:
• promotional campaigns
• bootstrapping liquidity
• other creative strategies

Because of that, I think the right place for guardrails is in UIs or management tools that can tailor safeguards to each user’s needs and level of expertise.

@wtogami @grubles , do you have any thoughts to add?

xmrk-btc:
> Yeah I think I lean more towards being hands-off here because, as Yusuke mentioned, there may be legitimate reasons we haven't thought of yet for node operators to set fee rates creatively.
> 
[...]

PR #428, opened 2026-03-29
multi: add premium rate sanity check warnings

Antisys:
## Summary

- Log warnings when operators set premium rates that may work against their own interest
- Rates are always accepted — warnings surface in logs for operators and UI tools to act on
- No API changes, no new dependencies

## Checks implemented (from #394)

- **Check 1**: SWAP_IN rate should be ≤ 0, SWAP_OUT rate should be ≥ 0
- **Check 2**: same-asset SWAP_IN + SWAP_OUT must be > 0 (prevents swap cycling profit)
- **Check 4**: LBTC_SWAP_IN + BTC_SWAP_OUT must be > 0 (prevents cross-asset arbitrage via free peg-in)

## Checks skipped

- **Check 3**: BTC_SWAP_IN + LBTC_SWAP_OUT > boltz_peg_out_rate — requires external Boltz rate data, left as TODO comment

## Design

Per maintainer feedback on #394, this is implemented as **log warnings** rather than hard blocks. Operators who intentionally use unusual rates (promotional campaigns, liquidity strategies) are not prevented from doing so. UI tools like peerswap-web can implement stricter guardrails on top of the validation functions exposed by the `premium` package.

## Test plan

- [x] Unit tests for each validation check (sign convention, same-asset sum, cross-asset sum)
- [x] Boundary value tests (±1, -999999+999999=0)
- [x] BTC and LBTC symmetry tests
- [x] Peer-specific rate lookup test
- [x] Order-of-operations end-to-end test (proves the hint in error messages works)
- [x] `go test ./...` passes
- [x] `make lint` passes with 0 issues

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/clightning/clightning_commands.go
+++ b/clightning/clightning_commands.go
@@ -1161,6 +1161,9 @@ func (c *UpdatePremiumRate) Call() (jrpc2.Result, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error creating premium rate: %v", err)
 	}
+	if err := c.cl.ps.ValidateRate(c.PeerID, rate); err != nil {
+		log.Infof("Premium rate warning: %v", err)
+	}
 	err = c.cl.ps.SetRate(context.Background(), c.PeerID, rate)
 	if err != nil {
 		return nil, fmt.Errorf("error setting premium rate: %v", err)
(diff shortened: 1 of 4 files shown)


YusukeShimizu:
Thanks for the PR. I understand the goal, but this validation falls outside the scope of the peerswap daemon.

This check acts as an operator policy / UX guardrail, not a strict protocol validity check. As noted in #394, safeguards like this should be handled by the caller application, not the core daemon.

Antisys:
Fair enough — the daemon should stay minimal and this is clearly a caller-side concern. Closing this.

I may look at contributing the validation to peerswap-web instead, where it fits more naturally.

Issue #435, opened 2026-05-13
peersync.NewAsset() silently drops poll messages from v5.0.0 peers due to case-sensitive asset parsing


ronballesteros:
### Summary
Nodes running the peersync-based peerswap (post-v5.0.0) silently drop all inbound poll messages from peers running v5.0.0, causing those peers to age the node off their active peer list and never reconnect.  

### Root Cause
`peersync.NewAsset()` performs a direct map lookup against stringToAsset, which only contains uppercase keys ("BTC", "LBTC"). However, v5.0.0 nodes advertise assets in lowercase ("btc", "lbtc") in their poll payloads. The lookup fails silently and the poll is discarded.   
                                
```
  // current — breaks on lowercase input
  func NewAsset(value string) (Asset, error) {
      if asset, ok := stringToAsset[value]; ok {                                    
          return asset, nil
      }                                                                             
      return AssetUnknown, fmt.Errorf("unsupported asset: %s", value)
  }
```

### Symptoms                                                                          
- Logs on the newer node show repeated: failed to parse poll message: invalid poll payload: unsupported asset: btc                          
- Remote v5.0.0 peer sees the node appear briefly then drop off their peer list within hours                                                                      
- Remote peer cannot reconnect without restarting their peerswap daemon

### Fix                                                       
                                                                                    
Normalize the input before lookup:                                                

```   
  func NewAsset(value string) (Asset, error) {                                      
      normalized := strings.ToUpper(strings.TrimSpace(value))
      if asset, ok := stringToAsset[normalized]; ok {
          return asset, nil
      }
      return AssetUnknown, fmt.Errorf("unsupported asset: %s", value)               
  }
```                                                             

### Impact                                                    
Any node that upgraded past v5.0.0 (where peersync replaced the old poll package) is silently incompatible with peers still on v5.0.0. The failure is not surfaced as an error to either side as peers just stop appearing as active.

<img width="700" height="300" alt="Image" src="https://github.com/user-attachments/assets/b75c848d-3186-4af7-82af-3f8be09729d1" />

YusukeShimizu:
Thanks for the report and fix.
I think this issue is resolved by #437.

PR #437, opened 2026-05-13
fix: peersync - case insensitive asset parsing

ronballesteros:
## Context

Related to https://github.com/ElementsProject/peerswap/issues/435
                                                                        
Nodes running the `peersync`-based daemon (post-v5.0.0) silently drop all inbound poll messages from peers still on v5.0.0, causing a one-way compatibility break.                                   

`NewAsset()` performs a direct map lookup against `stringToAsset`, which only contains uppercase keys (`"BTC"`, `"LBTC"`). v5.0.0 nodes advertise assets in lowercase in their poll payloads. The lookup fails, the poll is discarded, and `observedAt` never refreshes on the remote side. The remote peer ages the node off their active list within hours and cannot recover without restarting their daemon.                               
                                                                                    
### Logs on the node show: 
```                                                     
  failed to parse poll message: invalid poll payload: unsupported asset: btc
  failed to parse poll message: invalid poll payload: unsupported asset: lbtc
```

### Fix                                                                            
Normalize the input with `strings.ToUpper` before lookup so both uppercase and lowercase symbols are accepted.             

### Testing                                                                        
Verified manually against a v5.0.0 peer. Poll messages containing lowercase `"btc"`/`"lbtc"` are now parsed correctly and `observedAt` stays current.

### DIFF
--- a/peersync/peer.go
+++ b/peersync/peer.go
@@ -4,6 +4,7 @@ package peersync
 import (
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/elementsproject/peerswap/premium"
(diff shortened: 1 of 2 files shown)


YusukeShimizu (approved):
Approved. Thanks for the focused fix.  
I also checked the previous `misc_2` CI failure.
It was an unrelated integration-test/glightning shutdown race, and the rerun is green now.

PR #440, opened 2026-05-17
perf: batch-load swaps in ListPeers

### LINKED ISSUE #439: ListPeers performs N full swap database scans: O(n×m) scaling with peers and swap history
### Problem

`ListPeers` in `peerswaprpc/server.go` calls `p.swaps.ListSwapsByPeer(pubKey)` once per peer inside a loop:

```
  for _, pubKey := range pagePubKeys {
      swaps, err := p.swaps.ListSwapsByPeer(pubKey)
      ...
  }
```

 `ListSwapsByPeer` in `swap/store.go` performs a full BoltDB bucket scan on every call... deserializing every swap from JSON and filtering by
  peer in memory with no index:

```
  err = b.ForEach(func(k, v []byte) error {
      swap := &SwapStateMachine{}
      json.Unmarshal(v, swap)
      if swap.Data.PeerNodeId == peer {
          swaps = append(swaps, swap)
      }
      return nil
  })
```

This means every call to `ListPeers` performs N full database scans, where N is the number of peerswap peers.

### Impact

On our live node with 19 peerswap peers and a large swap history, ListPeers takes 37 seconds to respond and returns only 28KB of data. The data is tiny  all the time is spent scanning the swap database 19 times. This makes the psweb index page take 47+ seconds to load on every request.

More discussion here: https://github.com/Impa10r/peerswap-web/discussions/113

### Suggested Fix

Load all swaps once before the loop and partition by peer in memory:

```
  allSwaps, err := p.swaps.ListAllSwaps()
  if err != nil {
      return nil, err
  }

  swapsByPeer := make(map[string][]*swap.SwapStateMachine)
  for _, s := range allSwaps {
      if s.Data != nil {
          swapsByPeer[s.Data.PeerNodeId] = append(swapsByPeer[s.Data.PeerNodeId], s)
      }
  }

  for _, pubKey := range pagePubKeys {
      swaps := swapsByPeer[pubKey]
      view := format.BuildPeerView(pubKey, capability, swaps)
      ...
  }
```

This reduces the complexity from O(n×m) to O(m)... one scan regardless of peer count.

sisyphusse1-ops:
## Summary

Replace per-peer `ListSwapsByPeer()` calls with a single `ListSwaps()` call followed by in-memory partitioning by peer. This reduces database scans from N (one per peer) to 1, fixing the O(n×m) performance issue.

## Changes

- **peerswaprpc/server.go** (LND path): Batch-load all swaps before the peer loop and index by `PeerNodeId`.
- **clightning/clightning_commands.go** (CLN path): Same refactor for the Core Lightning implementation.

## How it works

Before: `ListPeers` called `ListSwapsByPeer(pubKey)` inside a per-peer loop. Each call performed a full BoltDB bucket scan deserializing every swap from JSON. With N peers and M total swaps, this was O(N×M).

After: One call to `ListSwaps()` loads all swaps, then a single pass partitions them into a `map[string][]*SwapStateMachine` keyed by peer. The per-peer loop does a simple map lookup. Complexity: O(M).

## Testing

- `go vet ./peerswaprpc/ ./clightning/` passes cleanly.
- Full `go build` / `go test` could not run due to disk constraints in CI environment, but the change is minimal and mechanical (no new logic, just reordering existing calls).

Closes #439

### DIFF
--- a/clightning/clightning_commands.go
+++ b/clightning/clightning_commands.go
@@ -574,15 +574,24 @@ func (l *ListPeers) Call() (jrpc2.Result, error) {
 		}
 	}
 
+	// Batch-load all swaps once and partition by peer to avoid O(n×m) scaling.
+	allSwaps, err := l.cl.swaps.ListSwaps()
+	if err != nil {
+		return nil, err
+	}
+	swapsByPeer := make(map[string][]*swap.SwapStateMachine)
+	for _, s := range allSwaps {
+		if s.Data != nil {
+			swapsByPeer[s.Data.PeerNodeId] = append(swapsByPeer[s.Data.PeerNodeId], s)
+		}
+	}
+
 	peerSwappers := []*peerswaprpc.PeerSwapPeer{}
 	for _, peer := range peers {
 		peerState, ok := compatible[peer.Id]
 		if ok {
 			capability := peerState.Capability()
-			swaps, err := l.cl.swaps.ListSwapsByPeer(peer.Id)
-			if err != nil {
-				return nil, err
-			}
+			swaps := swapsByPeer[peer.Id]
 
 			view := format.BuildPeerView(peer.Id, capability, swaps)
 			peerSwapPeer := peerswaprpc.NewPeerSwapPeerFromView(view)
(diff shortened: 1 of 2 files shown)


YusukeShimizu:
I checked for obvious implementation/panic/error-handling concerns and did not find anything blocking.

@ronballesteros, could you verify this branch on the live node from the original report?  
The key confirmation is whether `ListPeers` is no longer taking tens of seconds with your large swap history.

ronballesteros:
@YusukeShimizu - Looks like listpeers is snappy:

https://github.com/ElementsProject/peerswap/issues/439#issuecomment-4492956851

Issue #442, opened 2026-05-30
Rounding error in `satsToAmountString()` elementsrpcwallet.go


nullcount:
I found this bug when I tried to use peerswap RPC to manually send LBTC sats from the connected elementsd wallet to a third-party swap address (non-peerswap). That third-party swap failed due to insufficient amount sent to lock-up address.

Suppose I sent the sats via pythonGRPC like this:

```
SAT_AMT = 123,456,142
request = psrpc.SendToAddressRequest(address=address, sat_amount=SAT_AMT)
resp: psrpc.SendToAddressResponse = await s.ps_stub.LiquidSendToAddress(request)
``` 
The actual amount that gets sent to the address was: 123,456,100 (minus 42 sats in this example because it rounded down)

The source of the bug is this function:
```
// satsToAmountString returns the amount in btc from sats
func satsToAmountString(sats uint64) string {
	bitcoinAmt := float64(sats) / 100000000
	return fmt.Sprintf("%f", bitcoinAmt)
}
```
In Golang, the default precision of `"%f"` is 6-decimals.

The fix is simple:

```  
return fmt.Sprintf("%.8f", bitcoinAmt)   // 8 decimals = 1-sat precision
```
This bug might appear in other places, but I only checked this one since it caused the bug I was facing.

YusukeShimizu:
We should fix this with exact satoshi-to-decimal formatting.  
I’ll keep this issue open until the fix is merged.

PR #443, opened 2026-05-30
fix: use 8-decimals for sat-level string formats

nullcount:
fix for: #442


YusukeShimizu:
Thanks for opening this fix.
Would you be okay with us closing this PR in favor of #444?

PR #444, opened 2026-06-03
fix: format BTC amounts exactly via integer sat-to-decimal

### LINKED ISSUE #442: Rounding error in `satsToAmountString()` elementsrpcwallet.go

nullcount:
fixes: #442

### DIFF
--- a/misc_tests/liquid_test.go
+++ b/misc_tests/liquid_test.go
@@ -433,7 +433,7 @@ func Test_RpcWalletPreimage(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	err = testSetup.FaucetCli(testSetup.walletCli, 1)
+	err = testSetup.FaucetCli(testSetup.walletCli, 100_000_000)
 	if err != nil {
 		t.Fatalf("err funding wallet %v", err)
 	}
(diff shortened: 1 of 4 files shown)


YusukeShimizu (approved):
Thanks for the fix. The integer-based satoshi formatting looks like the right direction.

PR #446, opened 2026-06-07
Format wallet send amounts exactly

### LINKED ISSUE #442: Rounding error in `satsToAmountString()` elementsrpcwallet.go

ibobgunardi:
## Summary
- format satoshi amounts with integer arithmetic before calling elementsd
- preserve all 8 decimal places instead of relying on `%f` default precision
- add coverage for the reported amount plus small, whole-coin, and max-value cases

Fixes #442

## Tests
- go test ./wallet -run TestSatsToAmountString -count=1 -v
- go test ./wallet -count=1

### DIFF
--- a/wallet/elementsrpcwallet.go
+++ b/wallet/elementsrpcwallet.go
@@ -213,6 +213,7 @@ const (
 	// 1 kb = 1000 bytes
 	kb              = 1000
 	btcToSatoshiExp = 8
+	satsPerBitcoin  = 100000000
 )
 
 func (r *ElementsRpcWallet) GetFee(txSize int64) (uint64, error) {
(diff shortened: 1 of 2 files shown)


YusukeShimizu:
Thanks for opening this PR.

This fixes the same issue as #444, and we are going to proceed with #444 since it also updates the related BTC amount formatting paths in the test helpers and fee-rate formatting.

Issue #445, opened 2026-06-05
swaps don't work with lnd 0.21-rc3, because lnd removed SendToRoute RPC


xmrk-btc:
Swaps stopped working after upgrading lnd from 0.21.0-rc1 to 0.21.0-rc3. When trying to do a swap, I get an error:
`Swap canceled. Reason: SendToRouteSync() rpc error: code = Unimplemented desc = unknown method SendToRouteSync for service lnrpc.Lightning
`

SendToRoute RPC was removed in this commit: 
https://github.com/lightningnetwork/lnd/commit/2e728d3bb62436583ac2c8c7cd328668a6307c1d
It was already deprecated, so the problem can only be fixed in peerswap.

YusukeShimizu:
Thanks for reporting this. This is the same LND 0.21 compatibility issue now tracked in #448.
The failure reported here is still valid and will be addressed through #448.

PR #449, opened 2026-06-19
lnd: Support LND 0.21 payment RPCs

### LINKED ISSUE #448: LND Version 0.21.0-Beta Compatibility
# Incompatibility with LND 0.21.0-beta: peerswap still calls removed legacy payment RPCs (`SendToRouteSync` / `SendPaymentSync`)

## Summary

LND 0.21.0 **removed the deprecated payment RPCs** from the `lnrpc.Lightning` and `routerrpc` services (the removal was announced for 0.21). peerswap still calls two of them in production code, so swaps break against an LND 0.21 node.

The user-visible symptom is the one already reported in #445:

```
SendToRouteSync() rpc error: code = Unimplemented desc = unknown method SendToRouteSync for service lnrpc.Lightning
```

This issue extends #445 with a **full repository audit**, the exact call sites, and a concrete migration plan.

## Affected LND version

`lnd v0.21.0-beta` (broke between `0.21.0-rc1` and `0.21.0-rc3`).

Release notes: https://github.com/lightningnetwork/lnd/blob/master/docs/release-notes/release-notes-0.21.0.md

> Removed the deprecated payment RPCs and `outgoing_chan_id` field that were announced for removal in 0.21.

| Removed RPC | Replacement |
|---|---|
| `lnrpc.SendPaymentSync` / `lnrpc.SendPayment` (stream) | `routerrpc.SendPaymentV2` |
| `lnrpc.SendToRouteSync` / `lnrpc.SendToRoute` (stream) | `routerrpc.SendToRouteV2` |
| `routerrpc.SendPayment` / `SendToRoute` / `TrackPayment` (stream) | `...V2` |
| field `outgoing_chan_id` in `lnrpc.QueryRoutesRequest` and `routerrpc.SendPaymentRequest` | `outgoing_chan_ids` (repeated) |

## Incompatibility map

Two production call sites and one test-framework call site call removed RPCs. Everything else is already on the v2 APIs.

### 1. `probePayment` uses `SendToRouteSync` — root cause of #445 (critical)

[`lnd/client.go` ~L437](https://github.com/ElementsProject/peerswap/blob/master/lnd/client.go#L437):

```go
res2, err := l.lndClient.SendToRouteSync(context.Background(), &lnrpc.SendToRouteRequest{
    PaymentHash: pHash,
    Route:       route.GetRoute(),
})
if err != nil {
    return false, "", fmt.Errorf("SendToRouteSync() %w", err)
}
if !strings.Contains(res2.PaymentError, "IncorrectOrUnknownPaymentDetails") {
    log.Debugf("send pay would be failed. reason:%w", res2.PaymentError)
    return false, res2.PaymentError, nil
}
return true, "", nil
```

This runs on **every swap-out**: `swap.ProbePayment` → `swap/actions.go` (`CheckPremiumAmount`/probe) and `swap/service.go`. With LND 0.21 the probe fails immediately, so swaps are completely broken.

### 2. `PayInvoice` uses `SendPaymentSync`

[`lnd/client.go` ~L226](https://github.com/ElementsProject/peerswap/blob/master/lnd/client.go#L226):

```go
func (l *Client) PayInvoice(payreq string) (preImage string, err error) {
    payres, err := l.lndClient.SendPaymentSync(l.ctx, &lnrpc.SendRequest{PaymentRequest: payreq})
    if err != nil {
        return "", nil // pre-existing bug: swallows the error
    }
    return hex.EncodeToString(payres.PaymentPreimage), nil
}
```

`PayInvoice` is part of the `swap.LightningClient` interface (`swap/services.go`). It calls a removed RPC and will fail on 0.21. Note the pre-existing bug on the error branch (`return "", nil` instead of `return "", err`) — worth fixing in the same change.

### 3. Test framework uses `SendPaymentSync`

[`testframework/lnd.go` ~L575](https://github.com/ElementsProject/peerswap/blob/master/testframework/lnd.go#L575) (`LndNode.PayInvoice`). Not production, but it breaks the integration test suite against LND 0.21.

### Already compatible — no change needed

- `PayInvoiceViaChannel` (`lnd/client.go` ~L297) already uses `routerrpc.SendPaymentV2` with the plural `OutgoingChanIds`.
- `BuildRoute` calls using `OutgoingChanId` (`lnd/client.go` ~L422, `testframework/lnd.go` ~L477, `lnd/lnd_test.go` ~L629) are **not** affected: the field was only removed from `QueryRoutesRequest` and `SendPaymentRequest`, not from `routerrpc.BuildRouteRequest`.
- Payment watching uses `invoicesrpc.SubscribeSingleInvoice` (`lnd/paymentwatcher.go`) — unaffected.
- `lnd/lnd_test.go` ~L639 already demonstrates `SendToRouteV2` and can serve as the reference implementation.

## Proposed fix

**No `go.mod` bump required.** The current pin (`lnd v0.18.4-beta.rc1`) already ships the v2 client stubs (`SendToRouteV2`, `SendPaymentV2`). This is a purely server-side removal — only the client call sites need to change.

### Fix 1 — `probePayment` → `SendToRouteV2`

This is **not a pure rename**, because the response type changes:
- `SendToRouteSync` returned `*lnrpc.SendResponse` with a `PaymentError` string.
- `SendToRouteV2` returns `*lnrpc.HTLCAttempt`; an HTLC failure is reported in `attempt.Failure.Code` (`lnrpc.Failure_FailureCode`) and **does not come back as a gRPC error** (`err == nil`, `Failure != nil`) — confirmed by the existing test at `lnd/lnd_test.go:643-648`.

```go
attempt, err := l.routerClient.SendToRouteV2(ctx, &routerrpc.SendToRouteRequest{
    PaymentHash: pHash,
    Route:       route.GetRoute(),
})
if err != nil { // real RPC error
    return false, "", fmt.Errorf("SendToRouteV2() %w", err)
}
// A successful probe = receiver rejects with IncorrectOrUnknownPaymentDetails.
if attempt.GetFailure().GetCode() == lnrpc.Failure_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS {
    return true, "", nil
}
return false, attempt.GetFailure().GetCode().String(), nil
```

This replaces the old `strings.Contains(res2.PaymentError, "IncorrectOrUnknownPaymentDetails")` check, which no longer works once `PaymentError` is gone.

### Fix 2 — `PayInvoice` → `SendPaymentV2`

Replace with a `routerClient.SendPaymentV2` stream (set `TimeoutSeconds`, `MaxParts`), iterating `Recv()` until `Payment_SUCCEEDED` / `Payment_FAILED` — the same pattern already used by `PayInvoiceViaChannel`. Also fix the error branch to `return "", err`.

### Fix 3 — test framework

`testframework/lnd.go` (`LndNode.PayInvoice`): replace `SendPaymentSync` with `RouterClient.SendPaymentV2` (the node already exposes a router client, used by `BuildRoute`). Required so the integration suite passes on LND 0.21.

## Validation

1. `go build ./...` and `go vet ./...`.
2. Manual swap-out against an LND 0.21 node (exercises `probePayment` + `PayInvoiceViaChannel`).
3. Integration suite under `lnd/` and `test/` pointed at LND 0.21.

## References

- #445 (original report)
- LND 0.21.0 release notes (link above)
- LND commit removing the legacy payment RPCs

YusukeShimizu:
LND 0.21 removes the legacy payment RPCs that PeerSwap still used in the LND payment and probe paths. That makes swaps fail against upgraded LND nodes even though the repository still builds against the older Go module stubs, so this updates those flows to the supported router V2 APIs.

The branch also keeps the Nix runtime update that exposed the incompatibility and adds an early CI guard for the removed LND RPCs. That turns future automated flake updates into a clearer compatibility signal before the full daemon matrix has to surface the same class of server-side API drift.

Fixes #448.
Related #445.
Supersedes #434.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -8,11 +8,11 @@
         "rust-overlay": "rust-overlay"
       },
       "locked": {
-        "lastModified": 1777325693,
-        "narHash": "sha256-jFOEQwFDRVghCDFu0mybSLeTk9zWJSQW9clWFMkCa5A=",
+        "lastModified": 1781782487,
+        "narHash": "sha256-HUiiSSjhod2Mln/GIQAA56LB/kXXm2Pc5NdB731IBw0=",
         "owner": "Blockstream",
         "repo": "electrs",
-        "rev": "503b740cce2133fd07f451cbe93249a4e092b300",
+        "rev": "c7956023c3bf2cb3ac8076a4f65ebdf14b509513",
         "type": "github"
       },
       "original": {
@@ -38,11 +38,11 @@
     },
     "crane_2": {
       "locked": {
-        "lastModified": 1778106249,
-        "narHash": "sha256-cM/AuKy5tMhwOOQIbha8ZRRMHVfNf7cv2aljIw+qoCg=",
+        "lastModified": 1781825982,
+        "narHash": "sha256-SlXKwIRIhrOSAcTjCB3ftPLzJWZStQIPS7J1FlZPnKk=",
         "owner": "ipetkov",
         "repo": "crane",
-        "rev": "6d015ea29630b7ad2402841386da2cb617a470a7",
+        "rev": "469fd08d0bcf6926321fa973c6777fbc87785dd7",
         "type": "github"
       },
       "original": {
@@ -217,11 +217,11 @@
     },
     "nixpkgs_3": {
       "locked": {
-        "lastModified": 1778458615,
-        "narHash": "sha256-cY07EsdhBJ8tFXPzDYevgqxRev9ZLxFonuq9wmq5kwg=",
+        "lastModified": 1781607440,
+        "narHash": "sha256-rxO+uc/KFbSJp+pgyXRuAX6QlG9hJdnt0BXpEQRXY+U=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "c6e5ca3c836a5f4dd9af9f2c1fc1c38f0fac988a",
+        "rev": "3e41b24abd260e8f71dbe2f5737d24122f972158",
         "type": "github"
       },
       "original": {
@@ -328,11 +328,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1778555852,
-        "narHash": "sha256-55EmwooVAS4UpA0oWd5wilKPRqCiHD5BAej9QiNwheY=",
+        "lastModified": 1781752752,
+        "narHash": "sha256-kVG5tV9hddPviGAgqf9sGSuStvv+HAB9onfKqGptV0k=",
         "owner": "oxalica",
         "repo": "rust-overlay",
-        "rev": "f29b0f7a9f367e0056b716f8aa137cb41e784444",
+        "rev": "c06d86dabe5b92982b9d67acccb9990d58da3a0e",
         "type": "github"
       },
       "original": {
--- a/flake.nix
+++ b/flake.nix
@@ -1,7 +1,7 @@
 {
   inputs = {
     # Base packages for devShell.
-    # nixpkgs is pinned via `flake.lock` (currently rev f997fa0f94fb1ce55bccb97f60d41412ae8fde4c).
+    # nixpkgs is pinned via `flake.lock` (currently rev 3e41b24abd260e8f71dbe2f5737d24122f972158).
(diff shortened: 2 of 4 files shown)


YusukeShimizu:
Replaced by #450, which uses branch `lnd-021-runtime-compatibility` and squashes the change into one documented commit.

PR #457, opened 2026-08-29
test(clightning): fix total paid field name

zloglevel:
Rename the private `DummyPayerWaiter.totalPayed` test field to the grammatically correct `totalPaid` and update its increment site.

This touches test code only and preserves the accumulated amount and all assertions. `go test ./clightning` passes.


zloglevel:
@YusukeShimizu Could you please take a look when you have a chance? Thanks!

PR #458, opened 2026-09-07
Disable new L-BTC swaps following the Liquid incident

YusukeShimizu:
Disable new L-BTC swap requests for v7.0.1 following the [public Liquid Network security incident](https://x.com/Liquid_BTC/status/2096696272447218108).

Both local RPC requests and incoming peer requests are rejected in both directions, including forced requests. L-BTC is omitted from capability announcements. Bitcoin swaps retain their existing behavior.

Liquid backend initialization, persisted swaps, and recovery callbacks remain enabled, including when Bitcoin swaps are disabled. Recovery still depends on Liquid network availability and the recovered chain; this change does not guarantee recovery. The README includes the official statement and screenshot.

Also stop the capability handler from incorrectly logging recognized swap messages, including `SwapInAgreement` (42073), as unknown. The existing swap handler continues to process them.

Validation:
- Unit/race suite, builds, and diff lint pass locally; final post-commit checks are running.
- Tests cover rejected requests, retransmissions, and persisted CSV/cooperative recovery with Bitcoin disabled.
- Liquid integration cases now check suspension. A pinned v7.0.0 fixture exercises restarting an existing swap with the updated plugin.
- Live-node integration requires Linux and will be verified in CI. Warren will separately test the recovery-only configuration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ out/test-builds/pscli
 
 # # build output of tools
 tools/bin/
+out/test-builds/peerswap-v7.0.0
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,9 @@ ${TEST_BIN_DIR}/peerswap:
 	chmod a+x ${TEST_BIN_DIR}/peerswap
 
 # Test section. Has commands for local and ci testing.
+ifeq ($(RUN_INTEGRATION_TESTS),1)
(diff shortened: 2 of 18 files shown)


YusukeShimizu:
Replaced by #459 to use the branch name disable-new-liquid-swaps.

